PR #127, opened 2021-08-03
New specest method: Superlets

tensionhead:
Author Guidelines
-----------------
- [ ] Is the change set **< 400 lines**? - new method took sadly more lines to implement
- [ ] Is the code running on all supported platforms? - couldn't test on windows

Reviewer Checklist
------------------
- [x] Code layout
  - [x] No commented-out code?
- [ ] Are all docstrings complete and accurate? Yes but documentation fails to build


syncopy/specest/const_def.py
```diff
@@ -0,0 +1,116 @@
...
+#: output conversion of complex fourier coefficients
+spectralConversions = {"pow": lambda x: (x * np.conj(x)).real.astype(np.float32),
+                       "fourier": lambda x: x.astype(np.complex128),
+                       "abs": lambda x: (np.absolute(x)).real.astype(np.float32)}
+
+#: available outputs of :func:`~syncopy.freqanalysis`
+availableOutputs = tuple(spectralConversions.keys())
+
+#: available tapers of :func:`~syncopy.freqanalysis`
+availableTapers = ("hann", "dpss")
```
pantaray, line 26: Very nice idea! Re-reading the code I like this "constant definition" file approach more and more. Really elegant!

Just two comments: the `availableTapers/Outputs` etc. are used by Sphinx to render the available options in the HTML docu. Now unfortunately, Sphinx is not smart enough to handle imported tuples to generate these option lists. Since we only use the `available*` tuples for building the docu, I propose moving those back into `freqanalysis`. 
In addition, Sphinx is panicking about not being able to build the API docs, since the location of functions has changed. We need to update the corresponding rst sources. I can go ahead and make these changes (move the tuples back to `freqanalysis` and repair the docu) unless you have any objections, @tensionhead

pantaray (approved):
As discussed, merge this initial batch first, then we'll worry about the docu.

PR #152, opened 2021-11-17
FIX: Modify stacking logic in ComputationalRoutine

### LINKED ISSUE #130: inducer bug - spy.selectdata-
Thank you for taking the time to report an issue! We know that writing detailed error reports takes effort, and we appreciate you taking the time.

**Describe the bug**
image

image

image

**To Reproduce**
1. Load/Import dataset 
2. select only certain amount of channel with the fuction spy.selectdata(data, channel = [x,y,z])
3. Error message/erroneous result. 

**Thank you again for your time!**

pantaray:
- new mandatory positional argument in `initialize` of `ComputationalRoutine`:
  `out_dimord` provides the output object's dimensional labels. This is used
  to determine the correct stacking dimension for trials
- do not default to the first dimension as stacking dimension: instead use the
  new `stackingDim` variable (which is determined based on `out_dimord`, see above)
- closes #130

On branch comproutine_fix
 Changes to be committed:
	modified:   syncopy/connectivity/connectivity_analysis.py
	modified:   syncopy/datatype/methods/selectdata.py
	modified:   syncopy/shared/computational_routine.py
	modified:   syncopy/specest/freqanalysis.py
	modified:   syncopy/tests/spy_setup.py
	modified:   syncopy/tests/test_computationalroutine.py
	modified:   syncopy/tests/test_specest.py

Author Guidelines
-----------------

Reviewer Checklist
------------------
- [ ] Code layout
  - [ ] No commented-out code?

### DIFF
--- a/syncopy/shared/computational_routine.py
+++ b/syncopy/shared/computational_routine.py
@@ -26,7 +26,7 @@ if sys.platform == "win32":
 # Local imports
 from .tools import get_defaults
 from syncopy import __storage__, __acme__, __path__
-from syncopy.shared.errors import SPYValueError, SPYWarning, SPYParallelError
+from syncopy.shared.errors import SPYValueError, SPYTypeError, SPYParallelError, SPYWarning
 if __acme__:
     from acme import ParallelMap
     import dask.distributed as dd
@@ -226,7 +226,7 @@ class ComputationalRoutine(ABC):
         self._callMax = 10000
         self._callCount = 0
 
-    def initialize(self, data, chan_per_worker=None, keeptrials=True):
+    def initialize(self, data, out_dimord, chan_per_worker=None, keeptrials=True):
         """
         Perform dry-run of calculation to determine output shape
 
@@ -235,6 +235,8 @@ class ComputationalRoutine(ABC):
         data : syncopy data object
            Syncopy data object to be processed (has to be the same object
            that is passed to :meth:`compute` for the actual calculation).
+        out_dimord : list
+           Data dimension labels of output object
         chan_per_worker : None or int
            Number of channels to be processed by each worker (only relevant in
            case of concurrent processing). If `chan_per_worker` is `None` (default)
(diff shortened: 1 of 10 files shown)


syncopy/shared/computational_routine.py
```diff
@@ -291,17 +293,34 @@ def initialize(self, data, chan_per_worker=None, keeptrials=True):
             dtp_list.append(dtype)
             trials.append(trial)
 
+        # Determine trial stacking dimension and compute aggregate shape of output
+        if "time" in out_dimord:        # AnalogData + SpectralData + CrossSpectralData
```
tensionhead, line 297: Might be that we have to come back here and add more variables?! Would it be maybe better to define a 'stackingDim' in the respective DataClass? It anyways knows it's own dimord of course, and hence can also inform the CR about it? It looks like the only use of the `out_dimord` is anyways exactly this no?

syncopy/tests/test_specest.py
```diff
@@ -918,6 +918,7 @@ class TestWavelet():
                        "channels": range(0, int(nChannels / 2)),
                        "toilim": [-20, 60.8]}]
 
+    @skip_low_mem
```
tensionhead, line 921: Uh this magic is new to me :)

syncopy/shared/computational_routine.py
pantaray: Jupp, absolutley right. Godd point. We could add a `stackingDim` attribute to each of our classes - that might make things more transparent too. I would actually prefer this approach. Do you want me to go ahead and implement it?

syncopy/tests/test_specest.py
pantaray: :D Let's see if that helps w/testing pipeline stability on low-spec systems. Fingers crossed!

syncopy/shared/computational_routine.py
tensionhead: Yes please go ahead :smiley: !

pantaray: Done, please re-review

syncopy/shared/computational_routine.py
```diff
@@ -294,17 +294,12 @@ def initialize(self, data, out_dimord, chan_per_worker=None, keeptrials=True):
             trials.append(trial)
 
         # Determine trial stacking dimension and compute aggregate shape of output
-        if "time" in out_dimord:        # AnalogData + SpectralData + CrossSpectralData
```
tensionhead: Perfect, also directly looks better :)

tensionhead (approved):

tensionhead (approved):
That's it!

PR #154, opened 2021-11-22
Spy arithmetic

pantaray:
Author Guidelines
-----------------

**Note** tests have not been implemented yet. This is intended as Code Review PR for the used methodology to enable arithmetic operations with Syncopy objects. However, basal functionality tests have been carried out sequentially, concurrently on a local machine and via SLURM. 

Reviewer Checklist
------------------
- [x] Code layout
  - [x] No commented-out code?

### DIFF
--- a/syncopy/datatype/base_data.py
+++ b/syncopy/datatype/base_data.py
@@ -738,8 +738,21 @@ class BaseData(ABC):
                 shutil.rmtree(os.path.splitext(self.filename)[0],
                               ignore_errors=True)
 
+    # Support for basic arithmetic operations (no in-place computations supported yet)
     def __add__(self, other):
         return _process_operator(self, other, "+")
+
+    def __sub__(self, other):
+        return _process_operator(self, other, "-")
+
+    def __mul__(self, other):
+        return _process_operator(self, other, "*")
+
+    def __truediv__(self, other):
+        return _process_operator(self, other, "/")
+
+    def __pow__(self, other):
+        return _process_operator(self, other, "**")
 
     # Class "constructor"
     def __init__(self, filename=None, dimord=None, mode="r+", **kwargs):
--- a/syncopy/datatype/base_data.py
+++ b/syncopy/datatype/base_data.py
@@ -25,6 +25,7 @@ import scipy as sp
 
 # Local imports
 import syncopy as spy
+from .methods.arithmetic import _add
 from .methods.selectdata import selectdata
 from .methods.show import show
 from syncopy.shared.tools import StructDict
@@ -727,6 +728,9 @@ class BaseData(ABC):
                 os.unlink(self.filename)
                 shutil.rmtree(os.path.splitext(self.filename)[0],
                               ignore_errors=True)
+
+    def __add__(self, other):
+        return _add(self, other)
 
     # Class "constructor"
     def __init__(self, filename=None, dimord=None, mode="r+", **kwargs):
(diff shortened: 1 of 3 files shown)


syncopy/datatype/base_data.py
```diff
@@ -738,9 +738,22 @@ def __del__(self):
                 shutil.rmtree(os.path.splitext(self.filename)[0],
                               ignore_errors=True)
 
+    # Support for basic arithmetic operations (no in-place computations supported yet)
     def __add__(self, other):
         return _process_operator(self, other, "+")
 
+    def __sub__(self, other):
+        return _process_operator(self, other, "-")
```
tensionhead, line 749: Overloading with non-member functions probably isn't standard, yet here absolutely the right way.. also looks super clean :+1:

syncopy/datatype/methods/arithmetic.py
```diff
@@ -165,10 +171,64 @@ def _parse_input(obj1, obj2, operator):
...
+    else:
+        opSel = None
+    log_dct = {"operator": operator,
+               "base": baseObj.__class__.__name__,
+               "base selection": baseObj._selection,
+               "operand": operand.__class__.__name__,
+               "operand selection": opSel}
+
+    # Create output object
+    out = baseObj.__class__(dimord=baseObj.dimord)
```
tensionhead, line 199: Very neat! What happens in cases like `data1 = data1 + data2` with the underlying hdf5 container(s) of `data1`?

syncopy/datatype/methods/arithmetic.py
```diff
@@ -165,10 +171,64 @@ def _parse_input(obj1, obj2, operator):
...
+            parallel = True
+        except ValueError:
+            parallel = False
+
+    # Perform actual computation
+    opMethod = SpyArithmetic(operand_dat, operand_idxs, operation=operation,
+                             opres_type=opres_type)
+    opMethod.initialize(baseObj,
+                        out._stackingDim,
+                        chan_per_worker=None,
```
tensionhead, line 231: Arithmetic should be almost trivially parallelisable over channels no? However probably not really needed..

tensionhead (commented):
Looks all really good to me!

syncopy/datatype/base_data.py
pantaray: :D Thanks. Necessity and invention and all that

syncopy/datatype/methods/arithmetic.py
pantaray: Yupp, should be. The only thing that's sabotaging it is the possible existence of in-place selections in one (or both) objects. Then the channel parallelization setup becomes complicated. Definitely not impossible, but unless users request it, I thought we could table that one for later

syncopy/datatype/methods/arithmetic.py
pantaray: Good catch!!! Works out of the box sequentially, but breaks in parallel.

syncopy/datatype/methods/arithmetic.py
tensionhead: Agreed :+1:

syncopy/datatype/methods/arithmetic.py
pantaray: Correction - works in parallel, my tests were just wrong. Sorry for the confusion!

Issue #12, opened 2020-01-17
Refactor meta-function tests


pantaray:
We currently do not test in-place selection w/`freqanalysis` - this should be addressed w/one or two test-cases. 

In general, we might want to consider cooking up a template for writing tests for Syncopy meta-functions to have all our analysis routines undergo standardized functional-/unit-/regression-testing.

PR #158, opened 2021-11-24
Spy arithmetic

pantaray:
Author Guidelines
-----------------
- [ ] Is the change set **< 400 lines**? No due to docstrings
- [ ] Is the code running locally **and** on the ESI cluster? Tests require ACME patch, only ran locally for the time being
- [ ] Is the code running on all supported platforms? Tested on macOS and Linux

Reviewer Checklist
------------------
- [ ] Code layout
  - [ ] No commented-out code?

### DIFF
--- a/syncopy/shared/computational_routine.py
+++ b/syncopy/shared/computational_routine.py
@@ -8,13 +8,10 @@ import os
 import sys
 import psutil
 import h5py
-import time
 import numpy as np
 from itertools import chain
 from abc import ABC, abstractmethod
-from collections.abc import Sized
 from copy import copy
-from glob import glob
 from numpy.lib.format import open_memmap
 from tqdm.auto import tqdm
 if sys.platform == "win32":
@@ -284,7 +281,7 @@ class ComputationalRoutine(ABC):
         trials = []
         for tk, trialno in enumerate(self.trialList):
             trial = data._preview_trial(trialno)
-            trlArg = tuple(arg[tk] if isinstance(arg, Sized) and len(arg) == self.numTrials \
+            trlArg = tuple(arg[tk] if isinstance(arg, (list, tuple, np.ndarray)) and len(arg) == self.numTrials \
                 else arg for arg in self.argv)
             chunkShape, dtype = self.computeFunction(trial,
                                                      *trlArg,
@@ -306,7 +303,6 @@ class ComputationalRoutine(ABC):
         chk_arr = np.array(chk_list)
         chunkShape = tuple(chk_arr.max(axis=0))
         if np.unique(chk_arr[:, stackingDim]).size > 1 and not self.keeptrials:
-            import pdb; pdb.set_trace()
             err = "Averaging trials of unequal lengths in output currently not supported!"
             raise NotImplementedError(err)
         if np.any([dtp_list[0] != dtp for dtp in dtp_list]):
@@ -337,7 +333,7 @@ class ComputationalRoutine(ABC):
 
         # Allocate control variables
         trial = trials[0]
-        trlArg0 = tuple(arg[0] if isinstance(arg, Sized) and len(arg) == self.numTrials \
+        trlArg0 = tuple(arg[0] if isinstance(arg, (list, tuple, np.ndarray)) and len(arg) == self.numTrials \
             else arg for arg in self.argv)
         chunkShape0 = chk_arr[0, :]
         lyt = [slice(0, stop) for stop in chunkShape0]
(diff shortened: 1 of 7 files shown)


syncopy/shared/computational_routine.py
```diff
@@ -284,7 +281,7 @@ def initialize(self, data, out_stackingdim, chan_per_worker=None, keeptrials=Tru
         trials = []
         for tk, trialno in enumerate(self.trialList):
             trial = data._preview_trial(trialno)
-            trlArg = tuple(arg[tk] if isinstance(arg, Sized) and len(arg) == self.numTrials \
+            trlArg = tuple(arg[tk] if isinstance(arg, (list, tuple, np.ndarray)) and len(arg) == self.numTrials \
```
tensionhead, line 284: This mechanism of (only!) distributing args (and not kwargs) along trials probably can still be made more robust and transparent.. but this we can/should tackle later

syncopy/tests/test_continuousdata.py
```diff
@@ -778,6 +785,157 @@ def test_sd_dataselection(self):
...
+            # swapped dimord)
+            result = operation(dummy, dummy2)
+            for tk, trl in enumerate(result.trials):
+                assert np.array_equal(trl, operation(dummy.trials[tk], dummy2.trials[tk]))
+            result = operation(ymmud, ymmud2)
+            for tk, trl in enumerate(result.trials):
+                assert np.array_equal(trl, operation(ymmud.trials[tk], ymmud2.trials[tk]))
+
+            # Now the most complicated case: user-defined subset selections are present
+            for trialSel in trialSelections:
```
tensionhead, line 839: This looks a bit like over-engineering to me.. especially the taper part

syncopy/tests/test_continuousdata.py
```diff
@@ -778,6 +785,157 @@ def test_sd_dataselection(self):
...
+        ymmud = SpectralData(data=np.transpose(self.data, [3, 2, 1, 0]),
+                             trialdefinition=self.trl,
+                             samplerate=self.samplerate,
+                             taper=["TestTaper_0{}".format(k) for k in range(1, self.nt + 1)],
+                             dimord=SpectralData._defaultDimord[::-1])
+        dummy2 = SpectralData(data=self.data,
+                              trialdefinition=self.trl,
+                              samplerate=self.samplerate,
+                              taper=["TestTaper_0{}".format(k) for k in range(1, self.nt + 1)])
+        ymmud2 = SpectralData(data=np.transpose(self.data, [3, 2, 1, 0]),
```
tensionhead, line 822: That this works surprises me.. isn't `self.data` some hdf5 container?

syncopy/tests/test_continuousdata.py
```diff
@@ -77,6 +77,109 @@
     + list(zip(["foilim"] * len(foilimSelections), foilimSelections))
 
 
+# Local helper function for performing basic arithmetic tests
+def _base_op_tests(dummy, ymmud, dummy2, ymmud2, dummyC, operation):
```
tensionhead, line 81: As far as I get this, you somehow try to abstract these selection tests to be  at least operation agnostic?! Would be great to have sth like this for general frontend tests, as this looks suuper cumbersome and error prone to re-implement each time

tensionhead (approved):
I can merely leave some comments.. this part of the code base still remains mostly opaque to me :/

syncopy/shared/computational_routine.py
pantaray: Agreed - I think the current mechanism works well enough but it's desperately lacking documentation...

syncopy/tests/test_continuousdata.py
pantaray: Nope, in this case `self` refers to the class `TestSpectralData` which has an attribute `data`: https://github.com/esi-neuroscience/syncopy/blob/1896c31556e2da39aaf18766731368648e95b6f6/syncopy/tests/test_continuousdata.py#L631

syncopy/tests/test_continuousdata.py
pantaray: Yes, agreed. This is a long-standing request for our test, cf. #12 (closed this recently since I still don't know how to make this work tbh)

syncopy/tests/test_continuousdata.py
pantaray: Yupp, this really tests everything and the kitchen sink but at the expense of significant test runtime. We might want to consider simplifying the selection test cases here. Thinking about a clever but still exhaustive test setup...

PR #160, opened 2021-11-26
Padding

pantaray:
Author Guidelines
-----------------

Reviewer Checklist
------------------
- [ ] Code layout
  - [ ] No commented-out code?

### DIFF
--- a/syncopy/datatype/methods/padding.py
+++ b/syncopy/datatype/methods/padding.py
@@ -12,12 +12,16 @@ from syncopy.shared.computational_routine import ComputationalRoutine
 from syncopy.shared.kwarg_decorators import unwrap_io
 from syncopy.shared.parsers import data_parser, array_parser, scalar_parser
 from syncopy.shared.errors import SPYTypeError, SPYValueError, SPYWarning
+from syncopy.shared.kwarg_decorators import unwrap_cfg, unwrap_select, detect_parallel_client
 
 __all__ = ["padding"]
 
 
+@unwrap_cfg
+@unwrap_select
+@detect_parallel_client
 def padding(data, padtype, pad="absolute", padlength=None, prepadlength=None,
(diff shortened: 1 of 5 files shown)


tensionhead (commented):
So regarding the meta-functions accepting numpy arrays, what is the expected shape encoding?

tensionhead (changes requested):
So I naively did a `from syncopy.datatype import padding` and tried to pad the test data generated by `tdat = generate_artificial_data(inmemory=True, seed=1230, nTrials=5)` but failed. Reason is that there is no default `padtype` and the doc just says
`padtype : str` so as a newbie user I got no idea what to do now. I strongly suggest setting a default `padtype` (which 99% of users want) and extending the doc respectively!

pantaray:
> So I naively did a `from syncopy.datatype import padding` and tried to pad the test data generated by `tdat = generate_artificial_data(inmemory=True, seed=1230, nTrials=5)` but failed. Reason is that there is no default `padtype` and the doc just says `padtype : str` so as a newbie user I got no idea what to do now. I strongly suggest setting a default `padtype` (which 99% of users want) and extending the doc respectively!

We can introduce a default `padtype`, however, the doc is pretty clear here, that `padtype` is a mandatory argument without default as in the [FieldTrip template](https://www.fieldtriptoolbox.org/reference/ft_preproc_padding/ )

pantaray:
> So regarding the meta-functions accepting numpy arrays, what is the expected shape encoding?

So far, no expectations, really, since the decorator does not really know what the meta-function is up to. The offending decorator just assumed that every Syncopy meta function must accept (at least) one Syncopy object. If the positional args of the meta function did not contain any Syncopy object, an exception was raised. Now `padding` is the first example of a meta-function where this limitation seems too restrictive (and I guess more meta-functions like it might turn up in the future).

PR #168, opened 2021-12-02
Synthetic data + equalitiy operator + support for CrossSpectralData selections

pantaray:
Author Guidelines
-----------------

Reviewer Checklist
------------------
- [x] Code layout
  - [x] No commented-out code?


syncopy/connectivity/AV_compRoutines.py
```diff
@@ -177,7 +177,7 @@ def process_metadata(self, data, out):
             out.trialdefinition = trl
         else:
             out.trialdefinition = np.array([[0, 1, 0]])
-            
+
```
tensionhead, line 180: I probably have to find a way to visualize these spaces in my editor ;)

syncopy/connectivity/AV_compRoutines.py
```diff
@@ -289,14 +289,14 @@ def pre_check(self):
...
             act = f"DataSet contains {self.numTrials} trials"
             raise SPYValueError(legal=lgl, varname="data", actual=act)
-    
+
     def process_metadata(self, data, out):
-        
+
         # Get trialdef array + channels from source
         if data._selection is not None:
             chanSec = data._selection.channel
```
tensionhead, line 302: The datatype here still is `CrossSpectralData` no? How/why does a `._selection.channel` work then here?

syncopy/connectivity/ST_compRoutines.py
```diff
@@ -399,13 +399,15 @@ class ST_CrossCovariance(ComputationalRoutine):
...
+        # in time-axis, trial offsets etc. are bogus anyway: simply take max-sample
+        # counts / 2 to fit lags
         if data._selection is not None:
             chanSec = data._selection.channel
-            trl = data._selection.trialdefinition
+            trl = np.ceil(data._selection.trialdefinition / 2)
         else:
             chanSec = slice(None)
-            trl = data.trialdefinition
+            trl = np.ceil(data.trialdefinition / 2)
```
tensionhead, line 410: So that was the missing piece to get it working in the end?! Was this tested with even/odd number of time-series input samples?

syncopy/datatype/continuous_data.py
```diff
@@ -379,7 +377,6 @@ def __init__(self, data=None, channel=None, samplerate=None, **kwargs):
 
         self.channel = channel
         self.samplerate = samplerate     # use setter for error-checking
-        self.data = data
```
tensionhead, line 385: So this was/is already done in the `BaseData` class? Good catch :+1:

syncopy/datatype/base_data.py
```diff
@@ -354,21 +354,48 @@ def _set_dataset_property_with_list(self, inData, propertyName, ndim):
...
+        sround = 50
+        smax = 1000
+        srate = min(max(min(smax, tlen / 2) // sround * sround, 1) for tlen in trialLens)
+        t0 = -srate
+        msg = "Artificially generated trial-layout: trigger offset = {t0} sec, " +\
+            "samplerate = {srate} Hz (rounded to {sround} Hz with max of {smax} Hz)"
+        SPYWarning(msg.format(t0=t0/srate, srate=srate, sround=sround, smax=smax), caller="data")
+
+        # Use constructed quantities to set up trial layout matrix
+        accumSamples = np.cumsum(trialLens)
```
tensionhead, line 406: I don't wanna overcomplicate it.. still have to ask, wouldn't it be also somewhat easy to iterate over the individual numpy arrays and then just stack the respective trial-lengths such that also different trial lengths/nSamples could be realized? It's really nothing I think we would need right now, but if we wanna write tests with non-equal lengths trials I could imagine this to be very helpful as we can create more complicated data on the fly!

syncopy/datatype/base_data.py
```diff
@@ -337,8 +337,17 @@ def _set_dataset_property_with_dataset(self, inData, propertyName, ndim):
         setattr(self, "_" + propertyName, inData)
 
     def _set_dataset_property_with_list(self, inData, propertyName, ndim):
-        """
-        Coming soon...
+        """Set a dataset property with list of NumPy arrays
+
+        Parameters
+        ----------
+            inData : list
```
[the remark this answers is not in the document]
pantaray: Hm, you're right: `list` here really just serves as a container for NumPy arrays. So yes, it doesn't make a difference if the arrays are packed into a `tuple`, `set` or any other iterable. I just picked list since we can easily check for it in `_set_dataset_property` Do you

syncopy/datatype/base_data.py
```diff
@@ -354,21 +354,48 @@ def _set_dataset_property_with_list(self, inData, propertyName, ndim):
...
+                act = "NumPy arrays of different shape"
+                raise SPYValueError(legal=lgl, varname="data", actual=act)
+            trialLens = [np.nanmax(val[:, self.dimord.index("sample")]) for val in inData]
+
+        # Now the shaky stuff: use determined trial lengths to cook up a (completely
+        # fictional) samplerate: we aim for `smax` Hz and round down to `sround` Hz
+        nTrials = len(trialLens)
+        sround = 50
+        smax = 1000
+        srate = min(max(min(smax, tlen / 2) // sround * sround, 1) for tlen in trialLens)
```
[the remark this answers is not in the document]
pantaray: I had the same thought - I modified class inheritance in the constructors to make this happen, cf https://github.com/esi-neuroscience/syncopy/blob/f6c22ac52c3c77fa54a192e81444a189bc550efe/syncopy/tests/test_basedata.py#L251

syncopy/datatype/base_data.py
```diff
@@ -354,21 +354,48 @@ def _set_dataset_property_with_list(self, inData, propertyName, ndim):
...
-        # use stackingdim to build data-array!
-        if self.__class__.__name__ == "AnalogData":
-            pass # use np.h/vstack here
         else:
-            pass # allocate ndarray
-
 
+            # Ensure all arrays have shape `(N, 3)``
+            if any(val.shape[1] != 3 for val in inData):
+                lgl = "NumPy 2d-arrays with 3 columns"
```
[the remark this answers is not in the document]
pantaray: :D Nope, `SpikeData` are "sparse" in their representation. Each spike is encoded via its sample number, channel and associated unit. So the data has shape `nSpikes x 3`, cf https://github.com/esi-neuroscience/syncopy/blob/7478257669adf862397d90c2a94a780ab02a1c23/syncopy/datatype/discrete_data.py#L344 

And `EventData` use `nEvents x 2` as shape, cf https://github.com/esi-neuroscience/syncopy/blob/7478257669adf862397d90c2a94a780ab02a1c23/syncopy/datatype/discrete_data.py#L524

However, `EventData` are themselves used to define trials, so the list generation does not really make sense, and is therefore not considered here.

syncopy/datatype/base_data.py
pantaray: Yes, absolutely. I just wanted to keep it simple for the time being. But this whole trial stacking can be extended to trials of different lengths - it just requires some add'l index gymnastics in for loops.

syncopy/connectivity/AV_compRoutines.py
pantaray: Good catch. Oversight on my end - I assumed, that `data` is an `AnalogData` object (as in the `ST_` routines). For `CrossSpectralData` the above channel selection does not work. I'll patch this.

syncopy/connectivity/ST_compRoutines.py
pantaray: Ahem, "tested" as in "I tried it even/odd sample counts". But no rigorous tests are in place yet.

syncopy/datatype/base_data.py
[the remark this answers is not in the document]
tensionhead: Ah.. ok it's for another data type, sorry was a bit confusing as these lines appear in `base_data.py` :)

syncopy/datatype/base_data.py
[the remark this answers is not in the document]
tensionhead: Ok great, I understand this somewhat cryptic line that one can pass an explicit `samplerate=...` to the constructor :+1:

syncopy/datatype/base_data.py
tensionhead: Ok I created #170 to keep track of this :+1:

syncopy/datatype/base_data.py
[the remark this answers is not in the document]
tensionhead: I think one could check for iterables with sth like `bool(iter(my_container))`, but I guess it's not critical and lists are the go-to containers for such cases imho!

syncopy/connectivity/AV_compRoutines.py
pantaray: Fix has been pushed (cf 61f2ee3)

PR #181, opened 2022-01-07
CrossSpectralData Tests

pantaray:
Author Guidelines
-----------------
- [ ] Is the change set **< 400 lines**? almost

Reviewer Checklist
------------------
- [ ] Code layout
  - [ ] No commented-out code?

### DIFF
--- a/syncopy/datatype/continuous_data.py
+++ b/syncopy/datatype/continuous_data.py
@@ -655,7 +655,7 @@ class CrossSpectralData(ContinuousData):
         # if data exists but no user-defined channel labels, create them on the fly
         if self._channel_i is None and self._data is not None:
             nChannel = self.data.shape[self.dimord.index("channel_i")]
-            return np.array(["channel_i-" + str(i + 1).zfill(len(str(nChannel)))
+            return np.array(["channel" + str(i + 1).zfill(len(str(nChannel)))
                              for i in range(nChannel)])
 
         return self._channel_i
@@ -685,7 +685,7 @@ class CrossSpectralData(ContinuousData):
         # if data exists but no user-defined channel labels, create them on the fly
         if self._channel_j is None and self._data is not None:
             nChannel = self.data.shape[self.dimord.index("channel_j")]
-            return np.array(["channel_j-" + str(i + 1).zfill(len(str(nChannel)))
+            return np.array(["channel" + str(i + 1).zfill(len(str(nChannel)))
                              for i in range(nChannel)])
 
         return self._channel_j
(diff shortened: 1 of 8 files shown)


syncopy/datatype/continuous_data.py
```diff
@@ -655,7 +655,7 @@ def channel_i(self):
         # if data exists but no user-defined channel labels, create them on the fly
         if self._channel_i is None and self._data is not None:
             nChannel = self.data.shape[self.dimord.index("channel_i")]
-            return np.array(["channel_i-" + str(i + 1).zfill(len(str(nChannel)))
+            return np.array(["channel" + str(i + 1).zfill(len(str(nChannel)))
```
tensionhead, line 658: Yup makes more sense, as the dimord label `channel_i` is anyways there :+1:

syncopy/tests/test_continuousdata.py
```diff
@@ -935,3 +923,216 @@ def test_sd_parallel(self, testcluster):
...
+                                  dimord=CrossSpectralData._defaultDimord[::-1])
+        ymmud.trialdefinition = self.trl
+
+        for obj in [dummy, ymmud]:
+            idx = [slice(None)] * len(obj.dimord)
+            timeIdx = obj.dimord.index("time")
+            chanIdx = obj.dimord.index("channel_i")
+            chanJdx = obj.dimord.index("channel_j")
+            freqIdx = obj.dimord.index("freq")
+            for trialSel in trialSelections:
```
tensionhead, line 1049: Here `itertools.product` might really help..?

tensionhead (commented):
Nice work :+1:

syncopy/tests/test_continuousdata.py
```diff
@@ -935,3 +923,216 @@ def test_sd_parallel(self, testcluster):
...
+                for chaniSel in chanSelections:
+                    for chanjSel in chanSelections:
+                        for timeSel in timeSelections:
+                            for freqSel in freqSelections:
```
tensionhead, line 1053: Here `itertools.product` might reduce some for-loops..?!

pantaray: Hm, interesting idea. It would definitely reduce the number of for loops but I'm not sure it would add anything in terms of legibility or performance (the lion share of computational time is taken up by the actual superlet computation, I don't think inefficiencies due to nested for loops are a concern here). Since the code is functional and we're on a relatively tight schedule I vote for leaving it as it is.

PR #210, opened 2022-02-15
Ft importer

tensionhead:
Author Guidelines
-----------------
- [x] Is the change set **< 440 lines**?
- [ ] only local tests for now

Reviewer Checklist
------------------
~- [ ] Are testing routines present?~ no, just feature addition PR
- [ ] Do objects in the global package namespace perform proper parsing of their input? **No**
- [ ] Code layout
  - [x] No commented-out code?


syncopy/io/load_ft.py
```diff
@@ -0,0 +1,419 @@
...
+    if 'fsample' in h5Group:
+        AData.samplerate = h5Group['fsample'][0, 0]
+    else:
+        AData.samplerate = _infer_fsample(h5File[time_refs[0]])
+
+    # -- retrieve shape information --
+    nTrials = trl_refs.size
+
+    # peek in 1st trial to determine single trial shape
+    # we only support equal trial lengths at this stage
```
pantaray, line 210: Makes total sense, however, I think some check or a more "controlled" try/except writing block to `ADset` might be warranted in case the trials do not have equal length. In its current form the code will crash with some cryptic HDF5 index error exception if any trial is longer/shorter than the first one

syncopy/io/load_ft.py
```diff
@@ -0,0 +1,419 @@
...
+    itemsize = h5File[trl_refs[0]].dtype.itemsize
+    # in Mbyte
+    trl_size = itemsize * nSamples * nChannels / 1e6
+
+    # assumption: single trial fits into RAM
+    if trl_size > 0.4 * mem_use:
+        msg = f"\nSingle trial is at least 40% of the requested chache size of {mem_use}MB\n"
+        msg += f"Still trying to load {trl_size:.1f}MB trials.."
+        SPYWarning(msg)
+
```
pantaray, line 223: I understand the reasoning behind not erroring out if `trl_size > 0.4 * mem_use`, however, relying on a warning alone has the potential to crash a user's Python session in case a trial really does not fit into memory. Maybe a raising an exception if `trl_size > mem_use` might be safer?

syncopy/io/load_ft.py
```diff
@@ -0,0 +1,419 @@
...
+    ADset = h5FileOut.create_dataset("data",
+                                     dtype=np.float32,
+                                     shape=[nTotalSamples, nChannels])
+
+    pbar = tqdm(trl_refs, desc=f"{struct_name} - loading {nTrials} trials")
+    SampleCounter = 0   # trial stacking
+
+    # one swipe per trial
+    for tr in pbar:
+        ADset[SampleCounter:nSamples, :] = h5File[tr]
```
pantaray, line 238: Is `h5File[tr]` a HDF5 dataset? If so, it could be streamed block by block from disk.

syncopy/io/load_ft.py
```diff
@@ -0,0 +1,419 @@
...
+    offsets = []
+    # we need to look into the time vectors for each trial
+    for time_r in time_refs:
+        offsets.append(h5File[time_r][0, 0])
+    offsets = np.array(offsets)
+
+    trl_def = np.hstack([sampleinfo, offsets[:, None]])
+    AData.trialdefinition = trl_def
+
+    # each channel label is an integer array with shape (X, 1),
```
pantaray, line 258: strings saved as integer arrays, oh boy

syncopy/io/load_ft.py
```diff
@@ -0,0 +1,419 @@
...
+    # where `X` is the number of ascii encoded characters
+    channels = [''.join(map(chr, h5File[cr][:, 0])) for cr in chan_refs]
+    AData.channel = channels
+
+    # -- additional fields --
+
+    # this is the most experimental part
+    AData.info = {}
```
pantaray, line 266: Now I get the `info` attribute. I like the idea! And yes, this definitely requires parsing the contents of the HDF5 group more thoroughly before attaching it to the instance.

syncopy/io/load_ft.py
```diff
@@ -0,0 +1,419 @@
...
+        msg += f"Still trying to load {trl_size:.1f}MB trials.."
+        SPYWarning(msg)
+
+    # -- IO process --
+
+    # create new hdf5 dataset for our AnalogData
+    # with the default dimord ['time', 'channel']
+    h5FileOut = h5py.File(AData.filename, mode="w")
+    ADset = h5FileOut.create_dataset("data",
+                                     dtype=np.float32,
```
pantaray, line 230: Shouldn't this be `h5File[trl_refs[tk]].dtype` or something like that? Does matlab save everything as `float32`?

syncopy/io/load_ft.py
```diff
@@ -0,0 +1,419 @@
...
+    # check equal trial lengths
+    for trl in structure['trial']:
+
+        if trl.shape[-1] != nSamples:
+            lgl = 'Trials of equal lengths'
+            actual = 'Trials of unequal lengths'
+            raise SPYValueError(lgl, varname="load .mat", actual=actual)
+
+        # channel x sample ordering in FT
+        trials.append(trl.T.astype(np.float32))
```
pantaray, line 326: Same as above - can we be sure the mat file always contains 32bit floats?

pantaray (commented):
Nice work @tensionhead ! As a minor comment, I think the docstrings could be a bit easier to follow if the "structures" would be always called "data structures" or something like this. I got a bit confused with the terms "structure" and "field". In addition, I think `add_fields` suggest doing something to the mat file - maybe `include_fields` or `import_fields` or something similar might be more intuitive.

syncopy/io/load_ft.py
```diff
@@ -0,0 +1,419 @@
...
+                add_fields=None,
+                mem_use=2000):
+
+    '''
+    Imports raw time-series data from Field Trip
+    into potentially multiple `~syncopy.AnalogData` objects,
+    one for each structure found within the MAT-file.
+
+    For MAT-File < v7.3 the MAT-file gets loaded completely
+    into RAM, but its size should be capped at 2GB.
```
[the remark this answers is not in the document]
tensionhead: MATLAB caps it according to the document linked under `See Also` section!

syncopy/io/load_ft.py
```diff
@@ -0,0 +1,419 @@
...
+
+    '''
+
+    # Required fields for the ft_datatype_raw
+    req_fields_raw = ('time', 'trial', 'label')
+
+    version = _get_Matlab_version(filename)
+    msg = f"Reading MAT-File version {version} "
+    SPYInfo(msg)
+
```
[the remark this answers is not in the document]
tensionhead: yap true, I first wanted to get sth working.. all the input validation is missing completely. Shall I add this already to this PR, or do it later together with the tests @pantaray ?

syncopy/io/load_ft.py
tensionhead: Yes, wanted to discuss that with you in any case.. I am absolutely fine to be really strict here! Will include an exception for those cases

syncopy/io/load_ft.py
tensionhead: It's a dataset containing a single trial, so reading a single trial at once is the strategy we deemed fitting to the rest of the Syncopy architecture?!

syncopy/io/load_ft.py
tensionhead: Nopes, but this magically casts it into `float32` which is our preferred format :)

syncopy/io/load_ft.py
tensionhead: They actually have 64bit float, but I got no problems with the test dataset provided by Mohsen

syncopy/io/load_ft.py
tensionhead: in ASCII encodeing :/

syncopy/io/load_ft.py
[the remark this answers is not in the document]
pantaray: Good point - I think later with tests would be the natural entry point. Thanks!

syncopy/io/load_ft.py
pantaray: Awesome, thanks!

syncopy/io/load_ft.py
pantaray: Got it, my bad - you're right of course!

syncopy/io/load_ft.py
pantaray: Ah, okay. Maybe we should issue a message (not sure if a warning is actually needed here) that the data is type-cast to float32? Just in the (admittedly rare if even existing) case someone has float64 data that actually depends on 64bit precision.

syncopy/io/load_ft.py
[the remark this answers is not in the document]
pantaray: Nicely explained, thanks!

syncopy/io/load_ft.py
pantaray: Nicely done!

PR #214, opened 2022-02-21
Bugfix PR

pantaray:
Author Guidelines
-----------------
- [ ] Is the code running on all supported platforms? **only tested on Linux (workstation + cluster)**

Reviewer Checklist
------------------
- [ ] Code layout
  - [ ] No commented-out code?


syncopy/datatype/base_data.py
```diff
@@ -1685,6 +1685,9 @@ def trialdefinition(self, data):
                         trlTime = data._get_time([trlno], toilim=[-np.inf, np.inf])[0]
                         if isinstance(trlTime, list):
                             stop = np.max(trlTime)
+                            # Avoid creating empty arrays for "static" `SpectralData` objects
```
tensionhead, line 1688: so this adds more or less a bogus time-axis for trial-stacking of say Fourier-Spectra?

syncopy/datatype/methods/selectdata.py
```diff
@@ -302,9 +305,20 @@ def selectdata(data, trials=None, channels=None, channels_i=None, channels_j=Non
...
 
+    # Inform the user what's about to happen
+    fauxTrials = [data._preview_trial(trlno) for trlno in data._selection.trials]
+    fauxSizes = [np.prod(ftrl.shape)*ftrl.dtype.itemsize for ftrl in fauxTrials]
+    selectionSize = sum(fauxSizes) / 1024**2
+    sUnit = "MB"
+    if selectionSize > 1000:
+        selectionSize /= 1024
+        sUnit = "GB"
+    msg = "Copying {dsize:3.2f} {dunit:s} of data based on selection " +\
```
tensionhead, line 318: Very nice :+1:

syncopy/shared/parsers.py
```diff
@@ -340,9 +340,17 @@ def array_parser(var, varname="", ntype=None, hasinf=None, hasnan=None,
     scalar_parser : similar functionality for parsing numeric scalars
     """
 
-    # Make sure `var` is array-like and convert it to ndarray to simplify parsing
+    # Make sure `var` is array-like
     if not isinstance(var, (np.ndarray, list)):
         raise SPYTypeError(var, varname=varname, expected="array_like")
+
+    # "Exotic" arrays (str et al.) must contain only elements of the same type
```
tensionhead, line 347: Ok this works indeed nicely now, and also makes the `sequence_parser` for function arguments like `('ch1', 'ch2')` not obsolete, as the array parser really expects arrays, which is good!

syncopy/datatype/methods/selectdata.py
tensionhead, line 318: I just tried adding two AnalogData objects together, and did not receive this message, which I would have expected as this triggers an on disc copying no?!

tensionhead (changes requested):

syncopy/datatype/base_data.py
pantaray: Exactly, but only "static" (i.e., not time-frequency) spectra, which are one sample "long" which was not quantified correctly.

syncopy/datatype/methods/selectdata.py
pantaray: Hm, I see. This is actually expected since arithmetic ops do not trigger a selectdata copy: the arithmetics are `ComputationalRoutine` instances themselves, hence they perform the copying on their own. Do you want me to add a message like this `_perform_computation` as well?

tensionhead: I think this would be good no? So ppl see what happens if they start chugging data around.. as even a simply `my_adata / 3` can trigger GBs of I/O operations

pantaray: Included in latest commit

tensionhead: wow, even the operator appears in the new Info message, really cool :+1:

tensionhead (commented):
Looks all good to me!

tensionhead (commented):
last try..

PR #219, opened 2022-02-25
FIX: Return correct trialtimes for SpikeData

### LINKED ISSUE #218: SpikeData trialtime
When accessing the property 'trialtime' of an SpikeData variable, it yields a a generator per spike, instead of per trial. For example:

spike_times = np.array([0, 3, 7, 12, 13, 15, 21])
channels = np.array([0, 0, 1, 0, 2, 0, 0], dtype = int)
units = np.array([0, 1, 0, 1, 1, 0, 0], dtype = int)

`a = spy.SpikeData(data = np.vstack([spike_times, channels, units]).T, trialdefinition = np.array([[0,5,2], [6,15,3]]), dimord = ['sample', 'channel', 'unit'], samplerate=5)`

When doing 

`len(a.trialtime)` we get 7 (i.e., number of spikes, not of trials).

I have coded an alternative way in which this could be done, but I don't know if it's the most efficient one! Here it is:

```
@property
def trialtime(self):
    """list(:class:`numpy.ndarray`): trigger-relative sample times in s"""
    if self.samplerate is not None and self.sampleinfo is not None:
        return [(trl[:,0] - self._t0[tk] - self.sampleinfo[tk, 0])/self.samplerate for tk, trl in enumerate(self.trials)]
```

Instead of the current definition at line 181 (discrete_data.py).

**System Profile:**
 - OS: [Windows 10]
 - Please paste the output of the following command here 
   ```python
   import syncopy as spy; spy.AnalogData().log
   ```

>>> SyNCopy v. 0.20 <<< 

Created: Fri Feb 25 10:23:06 2022 

System Profile: 
3.8.12 (default, Oct 12 2021, 13:49:34) 
[GCC 7.5.0] 
ACME:  0.2rc3
Dask:  2021.10.0
NumPy: 1.21.5
SciPy: 1.7.1

--- LOG ---

|=== tlaieboriaa@esi-svhpc1: Fri Feb 25 11:10:49 2022 ===|

	__init__: created AnalogData object

Thank you!

tensionhead:
- return a list of arrays with length nTrials instead of a generator with length nSpikes
- fixes #218 with the solution provided by @atlaie

Changes to be committed:
	modified:   syncopy/datatype/base_data.py
	modified:   syncopy/datatype/discrete_data.py

Author Guidelines
-----------------
this is a tiny bugfix
- [ ] ~Was the code checked for memory leaks/performance bottlenecks?~
- [ ] ~Is the code running locally **and** on the ESI cluster?~
- [ ] ~Is the code running on all supported platforms?~

Reviewer Checklist
------------------
- [ ] Code layout
  - [ ] No commented-out code?

### DIFF
--- a/syncopy/datatype/base_data.py
+++ b/syncopy/datatype/base_data.py
@@ -593,6 +593,7 @@ class BaseData(ABC):
 
     @property
     def _t0(self):
+        """ These are the trigger offsets """
         if self._trialdefinition is not None:
             return self._trialdefinition[:, 2]
         else:
--- a/syncopy/datatype/discrete_data.py
+++ b/syncopy/datatype/discrete_data.py
@@ -182,9 +182,10 @@ class DiscreteData(BaseData, ABC):
     def trialtime(self):
         """list(:class:`numpy.ndarray`): trigger-relative sample times in s"""
         if self.samplerate is not None and self.sampleinfo is not None:
-            return [((t + self._t0[tk]) / self.samplerate \
-                    for t in range(0, int(self.sampleinfo[tk, 1] - self.sampleinfo[tk, 0]))) \
-                    for tk in self.trialid]
+            return [
+                (trl[:, 0] - self._t0[tk] - self.sampleinfo[tk, 0]) / self.samplerate
+                for tk, trl in enumerate(self.trials)
+            ]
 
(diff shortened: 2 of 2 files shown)


pantaray (approved):
Nicely done, thanks @atlaie !

PR #220, opened 2022-02-25
NEW: Support tapers other than dpss with additional parameters

### LINKED ISSUE #183: General tapers support taperopt
All backends which use a `taper` parameter also have a `taperopt={..}`, however the frontends don't make any use of this right now, except of course for the `dpss` multi tapering. Even though we allow in `const_def.availableTapers` a bunch which have additional parameters (e.g. `gaussian`, `kaiser`, ...), for these we get `TypeErrors` at the moment. 

It should be straightforward to allow a more general `taperopt` keyword in the respective frontends (`freqanalysis`, `connectivity`).

tensionhead:
- tapers like 'kaiser' need additional parameters which now can be
passed from the fronted via the new `taper_opt` parameter
- no changes in backend needed as here a `taper_opt` was already present
- resolves #183

On branch general_tapers
Changes to be committed:
	modified:   syncopy/nwanalysis/connectivity_analysis.py
	modified:   syncopy/shared/const_def.py
	modified:   syncopy/shared/input_processors.py
	modified:   syncopy/specest/freqanalysis.py

Author Guidelines
-----------------
The introduced changes can't effect any of those points

Reviewer Checklist
------------------
- [ ] Code layout
  - [ ] No commented-out code?

### DIFF
--- a/syncopy/nwanalysis/connectivity_analysis.py
+++ b/syncopy/nwanalysis/connectivity_analysis.py
@@ -36,8 +36,8 @@ availableMethods = ("coh", "corr", "granger")
 @detect_parallel_client
 def connectivityanalysis(data, method="coh", keeptrials=False, output="abs",
                          foi=None, foilim=None, pad_to_length=None,
-                         polyremoval=None, taper="hann", taper_opt=None, tapsmofrq=None,
-                         nTaper=None, out=None, **kwargs):
+                         polyremoval=None, tapsmofrq=None, nTaper=None,
+                         taper="hann", taper_opt=None, out=None, **kwargs):
 
     """
     Perform connectivity analysis of Syncopy :class:`~syncopy.AnalogData` objects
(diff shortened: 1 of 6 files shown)


syncopy/nwanalysis/connectivity_analysis.py
```diff
@@ -115,6 +115,10 @@ def connectivityanalysis(data, method="coh", keeptrials=False, output="abs",
     taper : str
         Only valid if `method` is `'coh'` or `'granger'`. Windowing function,
         one of :data:`~syncopy.specest.const_def.availableTapers`
+    taper_opt : dict or None
+        Dictionary with keys for additional taper parameters.
+        For example :func:`~scipy.signal.windows.kaiser` has
+        the additional parameter 'beta'.
```
pantaray, line 121: The `process_taper` routine is starting to get pretty complex with lots of if/else branching. If I understand correctly when utilizing DPSS tapers any user-provided `taper_opt` is ignored and instead `dpss_opt = {'NW': NW, 'Kmax': nTaper}` is used. I think this should be emphasized in the docstring.

syncopy/specest/freqanalysis.py
```diff
@@ -165,6 +165,10 @@ def freqanalysis(data, method='mtmfft', output='pow',
     taper : str
         Only valid if `method` is `'mtmfft'` or `'mtmconvol'`. Windowing function,
         one of :data:`~syncopy.shared.const_def.availableTapers` (see below).
+    taper_opt : dict or None
+        Dictionary with keys for additional taper parameters.
+        For example :func:`~scipy.signal.windows.kaiser` has
+        the additional parameter 'beta'.
```
pantaray, line 171: See comment in `connectivityanalysis`

pantaray (commented):
Looks good to me - I would just clarify that `taper_opt` doesn't do anything with DPSS tapers.

syncopy/nwanalysis/connectivity_analysis.py
tensionhead: Yep, but at least it's re-usable for both `specest` and `nwanalysis` frontends ;) I am also not super happy about the current structure, but as you can see here; https://github.com/esi-neuroscience/syncopy/blob/f83188af815fe93ac84d2e18f4462cc326557c1f/syncopy/shared/input_processors.py#L246 The users are getting a warning, and `dpss` is only configurable via our `tapsmofrq` (and the not recommenden `nTaper`) 
I will amend the docstring!

pantaray: Great, thank you! Other than that, I think we can merge right away. Thanks @tensionhead !

pantaray (approved):
Nice work @tensionhead Feel free to merge!

PR #234, opened 2022-03-11
Add least-square solution to wilson factorization

tensionhead:
Author Guidelines
-----------------

Reviewer Checklist
------------------
- [ ] Code layout
  - [ ] No commented-out code?


syncopy/nwanalysis/wilson_sf.py
```diff
@@ -59,17 +63,25 @@ def wilson_sf(CSD, nIter=100, rtol=1e-9):
...
 
-        psi_inv = np.linalg.inv(psi)
-        # the bracket of equation 3.1
-        g = psi_inv @ CSD @ psi_inv.conj().transpose(0, 2, 1)
+        if direct_inversion:
+            psi_inv = np.linalg.inv(psi)
+            # the bracket of equation 3.1
+            g = psi_inv @ CSD @ psi_inv.conj().transpose(0, 2, 1)
+        else:
+            for i in range(nFreq):
```
pantaray, line 75: I think we don't need the for loop: `numpy.linalg.lstsq` can process right hand sides `b` with shape `(M,M,K)`, right? One transpose of `CSD` should do it and we don't have to iterate over frequencies.

pantaray (commented):
Looks all good - I think we could get rid of the for loop in `wilson_sf` which should improve performance.

syncopy/nwanalysis/wilson_sf.py
tensionhead: No they actually can not, shapes are: 
```
a : (M, N) array_like
    "Coefficient" matrix.
b : {(M,), (M, K)} array_like
```

pantaray: I see, my bad - the coefficient matrix hast to be 2d, the rhs can be 3d. Never mind then, thanks for the clarification.

pantaray (approved):
Looks all good.

Issue #240, opened 2022-03-16
toi and toilim in `show()`


tensionhead:
The return type of a call to `show` is different depending on the keyword:

```python
# out1 is an array as expected
out1 = my_data.show(toilim=[1, 2])
# out2 is a list with one entry, the (correct afaik) array
out2 = my_data.show(toi=np.linspace(1, 2, 5))
```

tensionhead:
The same is true for the `foi/foilim` pair.. a `show(foi=[1,2,3])` returns a list with one entry

tensionhead:
Similarly, for a selection outside of the data bounds with `toilim` a list holding one empty array is returned: 
```python
mock_up = np.arange(10).reshape(5, 2)
ad = spy.AnalogData([mock_up])
# ad.time gives [array([-1.,  0.,  1.,  2.,  3.])]
out = ad.show(toilim=[4, 5])  # this is a list holding the empty array
```
Repeating the same with `toi` will throw a misleading error message:
```python
out = ad.show(toi=[4, 5])
```
Throws:
`TypeError: Indexing elements must be in increasing order`

Here an array parsing as it is being done for `foi/foilim` would be needed!

tensionhead:
So actually the same is true for `channel` selections..
```python
# out1 is a array
out1 = my_data.show(channel=0)
# out2 is a list
out2 = my_data.show(channel=[1,2])
```

tensionhead:
The inconsistent return behaviour (list or not a list) got addressed in 7557e4a, the out of bounds errors should be addressed elsewhere!

tensionhead:
Closed with #249

PR #249, opened 2022-03-28
FIX: Always return lists when showing multiple trials

### LINKED ISSUE #239: `show()` method on multiple trials
At the moment if a user does something like `my_data.show(trials=[0,3,17])` a single numpy array is returned, where all (3 in this example) selected trials are stacked together along the time-axis. It's probably much more useful to return an iterable which contains the individual arrays for the selected trials. As usual we also have to take care of some sort of 'auto-squeeze`, I would propose:
```python
out1 = my_data.show(trials=17)   # out1 is an array holding the single trial
out2 = my_data.show(trials=[0, 3, 17], average=True)   # again this is a single array 
out3 = my_data.show(trials=[0, 3, 17])   # here out3 should be an iterable of length 3
```
Any other real `dim` selection can always be returned as part of a single array (e.g. `foilim`, `toilim`, `channel`, ...), so ony trial selections can be returned as iterables in cases where this is meaningful as indicated above.

tensionhead:
- closes #239 and #240
- the out of bounds behaviour of toi (not toilim) selections
is another pair of shoes entirely (downstream of `show`), and should get addressed elsewhere

On branch show-fixes
Changes to be committed:
	modified:   syncopy/datatype/methods/show.py
	modified:   syncopy/tests/test_selectdata.py

Author Guidelines
-----------------
Only minor modifications shouldn't change anything below
- ~[ ] Was the code checked for memory leaks/performance bottlenecks?~
- ~[ ] Is the code running locally **and** on the ESI cluster?~
- ~[ ] Is the code running on all supported platforms?~

Reviewer Checklist
------------------
- [ ] Code layout
  - [ ] No commented-out code?

### DIFF
--- a/syncopy/datatype/methods/show.py
+++ b/syncopy/datatype/methods/show.py
@@ -121,7 +121,7 @@ def show(data, squeeze=True, **kwargs):
         msg = "".join(selectionTxt[txtMask])
         transform_out = np.squeeze
     else:
-        transform_out = lambda x : x
+        transform_out = lambda x: x
     SPYInfo("Showing{}".format(msg))
 
     # Use an object's `_preview_trial` method fetch required indexing tuples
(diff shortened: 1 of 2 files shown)


syncopy/datatype/methods/show.py
```diff
@@ -132,24 +132,43 @@ def show(data, squeeze=True, **kwargs):
...
-    if all(si == True for si in singleIdx):
-        return transform_out(data.data[tuple(returnIdx)])
+    # single trial selected
+    if len(idxList) == 1:
+        return transform_out(data.data[idxList[0]])
+    # return multiple trials as list
     else:
         return [transform_out(data.data[idx]) for idx in idxList]
+
+    # FIXME: remove for the same reason as above
```
pantaray, line 168: Based on recent discussions I think both commented-out `FIXME` code blocks can go.

pantaray (commented):
Looks all good to me - the added tests make a lot of sense as well. Unless you have any objections @tensionhead I think we can remove the `FIXME` code blocks and merge.

syncopy/datatype/methods/show.py
pantaray: Nice and clean, thanks @tensionhead !

pantaray (approved):
Good to go!

PR #253, opened 2022-03-31
Change _selection to selection

KatharineShapcott:
To match with AnalogData class. I get an error that _selection doesn't exist without this change.

Author Guidelines
-----------------

Reviewer Checklist
------------------
- [ ] Code layout
  - [ ] No commented-out code?

### DIFF
--- a/syncopy/preproc/preprocessing.py
+++ b/syncopy/preproc/preprocessing.py
@@ -122,11 +122,11 @@ def preprocessing(data,
 
     # if a subset selection is present
     # get sampleinfo and check for equidistancy
-    if data._selection is not None:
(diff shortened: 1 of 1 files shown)


tensionhead:
Hey @KatharineShapcott , good catch! Actually this got also captured by the Butterworth tests #246 , which are still under review by @joschaschmiedt  and did not get merged into dev unfortunately. There should be also at least one other bug which got already fixed in #246, maybe for testing the filter and to avoid duplicated bug fixes/merge conflicts could you maybe directly work in [preproc-tests](https://github.com/esi-neuroscience/syncopy/tree/preproc-tests)?!

PR #264, opened 2022-04-11
Preproc hilbert + rectification + FIRWS tests

tensionhead:
Author Guidelines
-----------------

Reviewer Checklist
------------------
- [ ] Code layout
  - [ ] No commented-out code?

### DIFF
--- a/syncopy/preproc/compRoutines.py
+++ b/syncopy/preproc/compRoutines.py
@@ -157,8 +157,6 @@ class Sinc_Filtering(ComputationalRoutine):
         if data.selection is not None:
             chanSec = data.selection.channel
             trl = data.selection.trialdefinition
-            for row in range(trl.shape[0]):
-                trl[row, :2] = [row, row + 1]
         else:
             chanSec = slice(None)
             trl = data.trialdefinition
--- a/syncopy/preproc/compRoutines.py
+++ b/syncopy/preproc/compRoutines.py
@@ -330,10 +330,6 @@ def rectify_cF(dat, noCompute=False, chunkShape=None):
     Consequently, this function does **not** perform any error checking and operates
     under the assumption that all inputs have been externally validated and cross-checked.
 
-    See also
-    --------
-    `Scipy butterworth documentation <https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.butter.html>`_
-
     """
 
     # operation does not change the shape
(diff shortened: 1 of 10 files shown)


syncopy/tests/test_preproc.py
```diff
@@ -138,26 +139,40 @@ def test_but_kwargs(self):
                       'order': 4}
             # only for firws
             if 'minphase' in direction:
+
```
pantaray, line 142: I think here the pytest-native fail-assertion would be more elegant than the `failed` try/except block:
```python
with pytest.raises(SPYValueError) as err:
   self.test_but_filter(**kwargs)
   assert "expected 'onepass'" in str(err)
```
This fails if `self.test_but_filter(**kwargs)` does not raise an exception, hence no need to incorporate `failed = True/False`

pantaray: Same for the remaining `failed = True` tests

syncopy/tests/test_preproc.py
```diff
@@ -206,6 +221,267 @@ def test_but_parallel(self, testcluster=None):
...
+                                       **kwargs)
+
+        # test rectification
+        filtered = call(rectify=False)
+        assert not np.all(filtered.trials[0] > 0)
+        rectified = call(rectify=True)
+        assert np.all(rectified.trials[0] > 0)
+
+        # test simultaneous call to hilbert and rectification
+        failed = True
```
pantaray, line 241: as above

syncopy/tests/test_preproc.py
```diff
@@ -206,6 +221,267 @@ def test_but_parallel(self, testcluster=None):
...
+        # test hilbert outputs
+        for output in preproc.hilbert_outputs:
+            htrafo = call(hilbert=output)
+            if output == 'complex':
+                assert np.all(np.imag(htrafo.trials[0]) != 0)
+            else:
+                assert np.all(np.imag(htrafo.trials[0]) == 0)
+
+        # test wrong hilbert parameter
+        failed = True
```
pantaray, line 259: as above

syncopy/tests/test_preproc.py
```diff
@@ -206,6 +221,267 @@ def test_but_parallel(self, testcluster=None):
...
+        for direction in preproc.availableDirections:
+            kwargs = {'direction': direction,
+                      'order': 200}
+            self.test_firws_filter(**kwargs)
+        for order in [-2, 220, 5.6]:
+            kwargs = {'direction': 'twopass',
+                      'order': order}
+
+            if order < 1 and isinstance(order, int):
+                failed = True
```
pantaray, line 375: as above

syncopy/tests/test_preproc.py
```diff
@@ -206,6 +221,267 @@ def test_but_parallel(self, testcluster=None):
...
+                                       **kwargs)
+
+        # test rectification
+        filtered = call(rectify=False)
+        assert not np.all(filtered.trials[0] > 0)
+        rectified = call(rectify=True)
+        assert np.all(rectified.trials[0] > 0)
+
+        # test simultaneous call to hilbert and rectification
+        failed = True
```
pantaray, line 459: as above

syncopy/tests/test_preproc.py
```diff
@@ -206,6 +221,267 @@ def test_but_parallel(self, testcluster=None):
...
+            test_method = getattr(self, test_name)
+            if 'firws_filter' in test_name:
+                # test parallelisation along channels
+                test_method(chan_per_worker=2)
+            else:
+                test_method()
+        client.close()
+        ppl.ion()
+
+    def test_firws_hilbert_rect(self):
```
pantaray, line 442: If I'm not mistaken this code is identical to `test_but_rect` with the exception of the very first `ppfunc` invocation. I think if we're committed to DRY we could unify the codes in a single `test_rect` function that loops over `filter_class='but'` and `filter_class='firws'`

pantaray (commented):
Nice feature addition and clever use of "nested" computeFunction calls.

syncopy/preproc/preprocessing.py
```diff
@@ -222,16 +246,46 @@ def preprocessing(data,
...
+
+        rectified = AnalogData(dimord=data.dimord)
+        rectCR = Rectify()
+        rectCR.initialize(filtered,
+                          data._stackingDim,
+                          chan_per_worker=kwargs.get("chan_per_worker"),
+                          keeptrials=True)
+        rectCR.compute(filtered, rectified,
+                       parallel=kwargs.get("parallel"),
+                       log_dict=log_dict)
```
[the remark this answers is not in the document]
tensionhead: True, I somehow assumed that magically such calls would be appended :D

syncopy/preproc/compRoutines.py
```diff
@@ -301,3 +299,176 @@ def process_metadata(self, data, out):
...
+    # Re-arrange array if necessary and get dimensional information
+    if timeAxis != 0:
+        dat = dat.T       # does not copy but creates view of `dat`
+    else:
+        dat = dat
+
+    # operation does not change the shape
+    # but may change the number format
+    outShape = dat.shape
+    fmt = np.complex64 if output == 'complex' else np.float32
```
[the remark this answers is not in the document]
tensionhead: Mmh.. so the input won't be complex, only the output if the user chooses to get the 'full' hilbert output. So the line you suggested will not work as intended?! Let's maybe have a broader discussion about this important issue elsewhere.

syncopy/preproc/preprocessing.py
```diff
@@ -222,16 +246,46 @@ def preprocessing(data,
...
+        return rectified
```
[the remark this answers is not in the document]
tensionhead: Absolutely, wasn't so sure what would happen _on disc_ after the `del`..

syncopy/tests/test_preproc.py
tensionhead: aah, yeah that's clearly the better way 👍

syncopy/tests/test_preproc.py
tensionhead: Yes I was thinking about this, however the default `order` parameter is so different for such cases, that I think for clarity maybe it's ok like this?

syncopy/preproc/preprocessing.py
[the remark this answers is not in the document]
tensionhead: ..which for me still does look like it, as none of our derived CRs do anything with the log

syncopy/preproc/compRoutines.py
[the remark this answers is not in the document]
pantaray: Yes, you're right, my bad. Let's discuss this in the standup.

syncopy/preproc/preprocessing.py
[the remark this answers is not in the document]
pantaray: The "destructor" (`__del__` in `BaseData`) removes HDF5 containers associated to an object if the containers live inside the temporary `__storage__` folder (`~/.spy` by default)

syncopy/tests/test_preproc.py
pantaray: OK, let's keep them separate.

syncopy/preproc/compRoutines.py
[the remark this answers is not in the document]
pantaray: Resolving this specific discussion and pointing to the general issue we've been talking about #265

pantaray:
Good to merge from my end!

PR #311, opened 2022-07-20
140 multiple outputs

dfsp-spirit:
Note: Despite the name, this does NOT implement multiple outputs for compute functions. It is only a step in that direction, and removes the hdr/memmap support.

Author Guidelines
-----------------

Reviewer Checklist
------------------
- [x] Code layout
  - [x] No commented-out code?


tensionhead:
As the removed memmap support never was exposed to the user (afaik), I think we don't need any changelog entry

tensionhead:
This got already reviewed as a sub branch, passing tests are enough for me here! Thx @dfsp-spirit for leading the way for streamlining the CRs :)

PR #315, opened 2022-07-22
303 auxiliary meta info container

tensionhead:
## New `.info` property for all Syncopy datatypes

- it's basically just a dict which gets saved as one of the `_infoFileProperties` into the json
- via `tools.SerializableDict` we hopefully discourage users from dumping huge arrays into this auxiliary dictionary

Author Guidelines
-----------------

Reviewer Checklist
------------------
- [x] Code layout
  - [x] No commented-out code?


dfsp-spirit (changes requested):
Looks good, the only thing that should be done is adding a check for non-serializable keys (see failing test I added to the branch).

tensionhead:
Ah.. ok so the relation of 'hashable - serializable' is probably above my current CS knowledge grade :grinning:

PR #326, opened 2022-08-05
325 upgrade preprocessing

### LINKED ISSUE #325: upgrade preprocessing
## Upgrade Preprocessing

Upon user feedback, we want to include into `preprocessing`:

- Z scoring (de-meaning and normalization by variance)
- standalone detrending (`polyremoval` only works atm when another CR is executed)

tensionhead:
## Standalone detrending and zscoring

As requested in #325, added standalone (no filtering) detrending and zscore normalization 

closes #325 

Author Guidelines
-----------------

Reviewer Checklist
------------------
- [x] Code layout
  - [x] No commented-out code?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
 - Added FOOOF method as a post-processing option for the freqanalysis method mtmfft.
 - new global `spy.copy()` function which copies entire Syncopy objects on disk
 - Added `.info` attribute for all data classes to store auxiliary meta information
+- Added `zscore` normalization to `preprocessing`
 
 
 ### CHANGED
@@ -21,6 +22,7 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
 - `spy.cleanup` now has exposed `interactive` parameter
 - removed keyword `deep` from `copy()`, all our copies are in fact deep
 - demeaning after tapering for granger analysis
+- detrending is now possible without filtering in `preprocessing`
 
 ### FIXED
 - `out.cfg` global side-effects (sorry again @kajal5888)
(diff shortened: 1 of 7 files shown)


dfsp-spirit:
Looks very good, thanks! 

My only concern is the slightly decreased coverage, which is mainly due to all the corner cases in the `cF`s that do not get hit in tests, it seems (e.g., `if timeAxis != 0:`, `if selection is not None`, ...). We test those in other tests and I do not see why we should re-test them for every cF, but still it looks bad in the cov report. What"s your thought on this?

I guess there are 2 1/2 ways to handle this:

1) get over the cov percentage and be done
2) mark lines as 'ignore coverage' with comment (like `# pragma: no cover`)
2.5) In some cases like the check for time axis, we could put that check into a function used by all cFs, then it would be called by some tests. I think this is not a great idea for the cF use cases though, re-designing the code for the coverage report seems wrong if it is not helpful beyond that, and in this case of trivial `if` checks, I would say an extra function reduces readability and clarity. In some *other* cases it may be suitable though.

dfsp-spirit (commented):
Looks very good, thanks! 

My only concern is the slightly decreased coverage, which is mainly due to all the corner cases in the `cF`s that do not get hit in tests, it seems (e.g., `if timeAxis != 0:`, `if selection is not None`, ...). We test those in other tests and I do not see why we should re-test them for every cF, but still it looks bad in the cov report. What"s your thought on this?

I guess there are 2 1/2 ways to handle this:

1) get over the cov percentage and be done
2) mark lines as 'ignore coverage' with comment (like `# pragma: no cover`)
3) In some cases like the check for time axis, we could put that check into a function used by all cFs, then it would be called by some tests (and thus be covered). I think this is not a great idea for the cF use cases though, re-designing the code for the coverage report seems wrong if it is not helpful beyond that, and in this case of trivial `if` checks, I would say an extra function reduces readability and clarity. In some *other* cases it may be suitable though.

tensionhead:
Thanks for starting this discussion @dfsp-spirit, indeed the coverage problem needs to be addressed! One way would be refactoring as you proposed under 3. I tried this now for the `/preproc` CRs with a general `propagate_metadata` and it already greatly reduced the number of copied code snippets, which now we don't have to test each time individually. Let's see what the merciless codecov bot tells us now..

PR #343, opened 2022-08-29
340 interface additional cf returns

### LINKED ISSUE #148: Refactoring process_metadata
Right now the following code-block gets copied for every CR that has an `out` datatype which has a continuous time-axis:
```python
        # Some index gymnastics to get trial begin/end "samples"                      
        if data._selection is not None:                                               
            chanSec = data._selection.channel                                         
            trl = data._selection.trialdefinition                                     
            for row in range(trl.shape[0]):                                           
                trl[row, :2] = [row, row + 1]                                         
        else:                                                                         
            chanSec = slice(None)                                                     
            time = np.arange(len(data.trials))                                        
            time = time.reshape((time.size, 1))                                       
            trl = np.hstack((time, time + 1,                                          
                             np.zeros((len(data.trials), 1)),                         
                             np.array(data.trialinfo)))                               
                                                                                      
        # Attach constructed trialdef-array (if even necessary)                       
        if self.keeptrials:                                                           
            out.trialdefinition = trl                                                 
        else:                                                                         
            out.trialdefinition = np.array([[0, 1, 0]])
```
Would be nice to bind that somewhere higher up in the class hierarchy or at least write an auxiliary function.

### LINKED ISSUE #340: Interface additional cF returns
Now that we finally have the possibility to get more return values from our cFs thanks to #321 we should make use of it:

### Granger

- the Wilson factorization returns a boolean `conv` indicating successful termination of the algorithm which might be of interest
- Potential regularization of the CSD also returns the `factor`

### Specest Methods

- everywhere where single trial `freq` axis appear, we could/should at least compare the hashes as in the end there is one global `.freq`

tensionhead:
## Make use of additional cF returns

- Granger now returns 4 additional metadata entries stored in `.info` documenting the factorization
- equality of frequency axis for mtmfft and csd get checked via hashes
- Further refactored `process_metadata` to reduce copied code and have one place where the properties get attached from input to output (`propagate_properties`)

closes #148
closes #340 

Author Guidelines
-----------------

Reviewer Checklist
------------------
- [x] Code layout
  - [x] No commented-out code?

### DIFF
--- a/syncopy/shared/computational_routine.py
+++ b/syncopy/shared/computational_routine.py
@@ -1019,15 +1019,14 @@ class ComputationalRoutine(ABC):
 
 # --- metadata helper functions ---
 
-def propagate_metadata(in_data, out_data):
+def propagate_properties(in_data, out_data, keeptrials=True):
 
     """
-    Propagating metadata/selections (channels, trials, time, ...)
+    Propagating data class properties (channels, trials, time, ...)
     from the input object `in_data` to the output
-    object `out_data` is the task of
-    concrete (overloaded) `process_metadata` implementations.
+    object `out_data` and respecting selections.
 
-    Depending on the concrete CR, different propagations are needed.
+    Depending on the CR in question, different propagations are needed.
     For example
 
         AnalogData -> CrossSpectralData (connectivityanalysis)
@@ -1038,12 +1037,12 @@ def propagate_metadata(in_data, out_data):
 
         AnalogData -> AnalogData (preprocessing)
 
-    directly propagates all (including selections)
-    attributes from the input to the output
-    (same channels, time and so on)
+    directly propagates the properties from the input to the output
+    (same channels, trialdefinition/time and so on)
 
     This function is a general approach to unify the
-    propagation / adaptation of all needed attributes.
+    propagation / manipulation of all needed properties
+    done in the `process_metadata` implementations.
 
     Parameters
     ----------
@@ -1054,10 +1053,12 @@ def propagate_metadata(in_data, out_data):
     """
 
     # instance checkers
-    is_AD = lambda data: isinstance(data, spy.AnalogData)
+    is_Analog = lambda data: isinstance(data, spy.AnalogData)
+    is_Spectral = lambda data: isinstance(data, spy.SpectralData)
+    is_CrossSpectral = lambda data: isinstance(data, spy.CrossSpectralData)
 
     # simplest case, direct propagation between two AnalogData objects
-    if is_AD(in_data) and is_AD(out_data):
+    if is_Analog(in_data) and is_Analog(out_data):
 
         # get channels and trial selections
         if in_data.selection is not None:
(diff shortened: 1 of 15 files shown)


syncopy/specest/compRoutines.py
```diff
@@ -200,31 +206,17 @@ class MultiTaperFFT(ComputationalRoutine):
...
-            trl = np.hstack((time, time + 1,
-                             np.zeros((len(data.trials), 1)),
-                             np.array(data.trialinfo)))
-
-        # Attach constructed trialdef-array (if even necessary)
-        if self.keeptrials:
-            out.trialdefinition = trl
-        else:
-            out.trialdefinition = np.array([[0, 1, 0]])
+        check_freq_hashes(metadata, out)
```
dfsp-spirit, line 211: good idea to put this into a function

syncopy/shared/computational_routine.py
```diff
@@ -1019,15 +1019,14 @@ def process_metadata(self, data, out):
 
 # --- metadata helper functions ---
 
-def propagate_metadata(in_data, out_data):
+def propagate_properties(in_data, out_data, keeptrials=True):
```
dfsp-spirit, line 1022: very nice, this reduces quite a bit code duplication

syncopy/tests/backend/test_conn.py
```diff
@@ -236,15 +234,13 @@ def test_regularization():
...
-    assert CN > cmax
-
     # --- regularize CSD ---
 
-    CSDreg, fac = regularize_csd(CSD, cond_max=cmax, nSteps=25)
+    cmax = 1e4
+    CSDreg, fac, iniCN = regularize_csd(CSD, cond_max=cmax, nSteps=25)
+    # get initial CSD condition number, which is way too large!
+    print(iniCN)
+    assert iniCN > cmax
```
dfsp-spirit, line 243: would be nice to use 2nd arg of assert statement to print better error here, like

   assert iniCN > cmax, f"some message about var {var} here"

dfsp-spirit:
Looks good,

only 2 things:

* the changelog should be updated
* do you want to document the new Granger return values in the docstrings of the frontend, so that the users know about them?

syncopy/shared/computational_routine.py
tensionhead: As discussed, it's still not perfect imho.. I could also think of putting those property mappings into the respective module `compRoutines.py`, however there is already quite some crossover between `freqanalysis` and `connectivityanalysis` for example.

syncopy/tests/backend/test_conn.py
tensionhead: yes, I've seen this only in your code now.. did not know this is possible, but good idea :+1:

tensionhead: puh, there was definitely something off in any case.. I corrected it, still not sure how often ill conditioned CSDs appear in the wild :sweat_smile: 
But at least now we know at least _aposteriori_ what the factorization/regularization did thanks to the additional cF returns :)

tensionhead:
About your 2nd point regarding the docstring of the frontend, I now think we should programmatically fix/define the key's which are propagated from the CR/cF to `.info`. Then also Sphinx can pick them up easily via `:data:`syncopy.nwanalysis.....`, but for now I 'hardcoded' the `.info` keys into the frontend doc.

tensionhead:
I also added the changelog entries for the additional fooof returns, which are also new right :) ?! Feel free to edit/change!

PR #355, opened 2022-09-21
344 rng

dfsp-spirit:
Adds option to get reproducible data from all synth_data functions by setting a seed.

* This is implemented by adding a new `seed` parameter to all functions.
* Special care has been taken with the `@collect_trials` wrapper: typically you do **not** want all trials to be identical, even if you want your data to be reproducible (e.g., when you average white noise from 100 trials, you do not want to get a clear signal). To achieve this, the boolean parameter `seed_per_trial` has been added to the `@collect_trials` wrapper, and is on by default. When on, it ensures that you get reproducible data that *differs* between trials. If you want identical trials, set `seed_per_trial=False`.
    - Obviously, if you don't reproducible data at all, don't pass a seed (or use `None` as the seed).

Author Guidelines
-----------------

Reviewer Checklist
------------------
- [ ] Code layout
  - [ ] No commented-out code?

### DIFF
--- a/syncopy/tests/synth_data.py
+++ b/syncopy/tests/synth_data.py
@@ -233,7 +233,7 @@ def AR2_network(AdjMat=None, nSamples=1000, alphas=[0.55, -0.8], seed=None):
 
     for i in range(2, nSamples):
         sol[i, :] = (DiagMat + AdjMat.T) @ sol[i - 1, :] + alpha2 * sol[i - 2, :]
-        sol[i, :] += rng.random((nChannels))
+        sol[i, :] += rng.normal(size=(nChannels))
 
     return sol
 
--- a/syncopy/tests/synth_data.py
+++ b/syncopy/tests/synth_data.py
@@ -231,11 +231,11 @@ def AR2_network(AdjMat=None, nSamples=1000, alphas=[0.55, -0.8], seed=None):
     sol = np.zeros((nSamples, nChannels))
     # pick the 1st values at random
     rng = np.random.default_rng(seed)
-    sol[:2, :] = rng.randn(2, nChannels)
+    sol[:2, :] = rng.random((2, nChannels))
 
     for i in range(2, nSamples):
         sol[i, :] = (DiagMat + AdjMat.T) @ sol[i - 1, :] + alpha2 * sol[i - 2, :]
-        sol[i, :] += rng.randn(nChannels)
+        sol[i, :] += rng.random((nChannels))
 
     return sol
 
--- a/syncopy/tests/synth_data.py
+++ b/syncopy/tests/synth_data.py
@@ -67,7 +67,7 @@ def collect_trials(trial_generator):
             for trial_idx in range(nTrials):
                 if 'seed' in signature(trial_generator).parameters.keys():
                     if seed_array is not None:
-                        tg_kwargs['seed'] = seed[trial_idx]
+                        tg_kwargs['seed'] = seed_array[trial_idx]
                     else:
                         tg_kwargs['seed'] = seed
                 trl_arr = trial_generator(**tg_kwargs)
--- a/syncopy/tests/synth_data.py
+++ b/syncopy/tests/synth_data.py
@@ -87,7 +87,7 @@ def white_noise(nSamples=1000, nChannels=2, seed=None):
     Plain white noise with unity standard deviation
     """
     rng = np.random.default_rng(seed)
-    return rng.random((nSamples, nChannels))
+    return rng.normal(size=(nSamples, nChannels))
 
 
 @collect_trials
--- a/syncopy/tests/synth_data.py
+++ b/syncopy/tests/synth_data.py
@@ -370,22 +370,24 @@ def poisson_noise(nTrials=10,
     """
 
     # uniform random weights
-    def get_rdm_weights(size):
-        pvec = np.random.uniform(size=size)
+    def get_rdm_weights(size, seed=seed):
+        rng = np.random.default_rng(seed)
+        pvec = rng.uniform(size=size)
         return pvec / pvec.sum()
 
     # total length of all trials combined
+    rng = np.random.default_rng(seed)
     T_max = int(nSpikes / intensity)
 
-    spike_samples = np.sort(random.sample(range(T_max), nSpikes))
-    channels = np.random.choice(
+    spike_samples = np.sort(rng.random(size=(range(T_max), nSpikes)))
+    channels = rng.choice(
         np.arange(nChannels), p=get_rdm_weights(nChannels),
         size=nSpikes, replace=True
     )
 
     uvec = np.arange(nUnits)
     pvec = get_rdm_weights(nUnits)
-    units = np.random.choice(uvec, p=pvec, size=nSpikes, replace=True)
+    units = rng.choice(uvec, p=pvec, size=nSpikes, replace=True)
 
     # originally fixed trial size
     step = T_max // nTrials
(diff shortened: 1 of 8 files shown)


syncopy/tests/synth_data.py
```diff
@@ -29,26 +30,47 @@ def collect_trials(trial_generator):
...
     just returns the output of the trial generating function
     directly, so a single trial :class:`numpy.ndarray`.
     """
 
     @functools.wraps(trial_generator)
-    def wrapper_synth(nTrials=None, samplerate=1000, **tg_kwargs):
+    def wrapper_synth(nTrials=None, samplerate=1000, seed=None, seed_per_trial=True, **tg_kwargs):
+
+        use_seed_per_trial = False
+        if nTrials is not None and seed is not None and seed_per_trial:  # Use the single seed to create one seed per trial.
```
tensionhead, line 49: Uff, that's a bit convoluted the logic here no? Or maybe I can't really grasp the situation..

syncopy/tests/synth_data.py
```diff
@@ -29,26 +30,47 @@ def collect_trials(trial_generator):
...
+    (and it will differ between trials).
+
     The default `nTrials=None` is the identity wrapper and
     just returns the output of the trial generating function
     directly, so a single trial :class:`numpy.ndarray`.
     """
 
     @functools.wraps(trial_generator)
-    def wrapper_synth(nTrials=None, samplerate=1000, **tg_kwargs):
+    def wrapper_synth(nTrials=None, samplerate=1000, seed=None, seed_per_trial=True, **tg_kwargs):
```
tensionhead, line 46: What about setting a default parameter for `seed`, such that we automatically inject a stable seed into all our current and future tests?

syncopy/tests/synth_data.py
```diff
@@ -29,26 +30,47 @@ def collect_trials(trial_generator):
...
+
+        use_seed_per_trial = False
+        if nTrials is not None and seed is not None and seed_per_trial:  # Use the single seed to create one seed per trial.
+            rng = np.random.default_rng(seed)
+            seed = rng.integers(1000000, size=nTrials)
```
tensionhead, line 51: wait, so `seed` gets redefined? What you probably mean are the by-trial seeds?!

Edit: Indeed, I think I would prefer to have a new variable ala `trl_seeds`, what do you think?

syncopy/tests/synth_data.py
```diff
@@ -29,26 +30,47 @@ def collect_trials(trial_generator):
...
+            use_seed_per_trial = True
 
         # append samplerate parameter if also needed by the generator
         if 'samplerate' in signature(trial_generator).parameters.keys():
             tg_kwargs['samplerate'] = samplerate
 
-        # do nothing
+        # do nothing (may pass on the scalar seed if the function supports it)
         if nTrials is None:
+            if 'seed' in signature(trial_generator).parameters.keys():
```
tensionhead, line 60: Ok I think this is elegant, akin to the treatment of `samplerate` we probably should somehow enforce or at least document this (expected) signature of the trial generators? I think we anyways must overhaul some other things I basically spontaneously cooked up, so we can also address this in another PR.

syncopy/tests/synth_data.py
```diff
@@ -29,26 +30,47 @@ def collect_trials(trial_generator):
...
         else:
             trl_list = []
 
-            for _ in range(nTrials):
+            for trial_idx in range(nTrials):
+                if 'seed' in signature(trial_generator).parameters.keys():
+                    if use_seed_per_trial:
+                        tg_kwargs['seed'] = seed[trial_idx]
+                    else:
+                        tg_kwargs['seed'] = seed
```
tensionhead, line 72: with this we just generate identical trial copies right? I don't think there is any real use case?!

syncopy/tests/synth_data.py
```diff
@@ -29,26 +30,47 @@ def collect_trials(trial_generator):
...
+                print(f"[collect_trials] use_seed_per_trial={use_seed_per_trial}. Calling {trial_generator.__name__} with kwargs: {tg_kwargs}")
```
tensionhead, line 73: Maybe use a `SPYInfo` here?

syncopy/tests/synth_data.py
```diff
@@ -101,7 +120,8 @@ def phase_diffusion(freq,
                     fs=1000,
```
tensionhead, line 120: This should have been `samplerate` right from the 1st implementation, before merge I would push a little commit adressing this

syncopy/tests/synth_data.py
```diff
@@ -202,29 +233,27 @@ def AR2_network(AdjMat=None, nSamples=1000, alphas=[0.55, -0.8]):
 
     sol = np.zeros((nSamples, nChannels))
     # pick the 1st values at random
-    sol[:2, :] = np.random.randn(2, nChannels)
+    rng = np.random.default_rng(seed)
```
tensionhead, line 236: Just wondering, there is no way to solve that more generally? However something like a global `rng` can't be it, as then the order of calls destroys reproducibility?

tensionhead (commented):
Great work @dfsp-spirit :rocket: Moving to the recommended `rng` API of NumPy makes a lot of sense, this should safe us quite some headaches from randomly failing tests :slightly_smiling_face:

syncopy/tests/synth_data.py
dfsp-spirit: If 'use_seed_per_trial' is true, we want fixed, but different seeds for each trial. If we would use the same seed in each trial for things like white noise, averaging over many white noise channels will NOT remove the noise.

So we need a way to create different but fixed seeds for all the trials. My solution is to generate them from the given seed.

Another option is to have the user pass as many seeds as there are trials. I had implemented this solution first, but it turned out that it leads to a lot of replicated and pointless code when we call these functions. Instead of something like:

    data = generate_whatever(,,, seed=42)

you get:

    data = generate_whatever(,,, seed=np.random.default_rng(seed).integers(1000000, size=nTrials))

in a lot of places.

syncopy/tests/synth_data.py
dfsp-spirit: Yes, with `use_seed_per_trial=False` you get the old behaviour (before this PR) that all trials are equal. Plus you can, optionally, make sure they are equal and fixed (by giving a seed that is different from `None`).

syncopy/tests/synth_data.py
dfsp-spirit: Yes, using the global RNG is not considered good practice anymore according to the (now accepted) [NEP 19](https://numpy.org/neps/nep-0019-rng-policy.html). Quoting :

_The preferred best practice for getting reproducible pseudorandom numbers is to instantiate a generator object with a   seed and pass it around. The implicit global RandomState behind the numpy.random.* convenience functions can cause problems, especially when threads or other forms of concurrency are involved. Global state is always problematic. We categorically recommend avoiding using the convenience functions when reproducibility is involved._

tensionhead: Ok, but what I am reading is *"pass it around"* and not re-instantiate in every function? Actually this passing around is probably what I meant by global, so not a global state.. but a global `rng` instance.

syncopy/tests/synth_data.py
tensionhead: Mmh, so the old behaviour was not equal trials right? Otherwise, many of our tests would make little sense.. But completely equal trials make only very little sense imho. We can leave it like that, but I expect that we'll never use this code branch

syncopy/tests/synth_data.py
tensionhead: Yes yes, I got this.. generating further by-trial-seeds from an initial seed is a good idea! The only nitpicky thing I had, was the redefinition of the `seed` name, it makes the code harder to read as it potentially hides exactly that sampling strategy (first `seed` is a number, than with `seed_per_trial = True` it is an array).

dfsp-spirit: agree, changed in 85300df

syncopy/tests/synth_data.py
dfsp-spirit: I added a new var to to make it less confusing.

syncopy/tests/synth_data.py
dfsp-spirit: done

syncopy/tests/synth_data.py
dfsp-spirit: I just deleted it.

tensionhead (approved):
Ok perfect, thx again @dfsp-spirit! I also really like our global seed of `42` :grinning: As mentioned before, I will replace the old `fs` with `samplerate` where this applies and then merge!

PR #358, opened 2022-10-07
Fix: tags loaded precisely

KatharineShapcott:
Previously part matches of tags were also loaded

Author Guidelines
-----------------

Reviewer Checklist
------------------
- [ ] Code layout
  - [ ] No commented-out code?


KatharineShapcott:
If you want the previous behaviour then you can add a star into the tag name itself e.g. `tag='lfp*'`

tensionhead:
@KatharineShapcott thx for your contributions! Could you please reopen this PR to merge into `dev` not `master`? Thx 🙂

tensionhead:
One more comment: please branch out from `dev`, not from `master` when you want to commit your changes. We should probably make this rule more visible to contributors. I think in this case it is fine, so just go ahead 🚀

KatharineShapcott:
Sorry I usually do, mistake thanks for fixing it! Still trying to figure out why this test is failing atm.

tensionhead:
Hey @KatharineShapcott you want to pick up in this again? Otherwise I could also have a look into the `SpyIO Error`, as I right now don't see how this could be connected to the changes you made?!

KatharineShapcott:
I suspect it is a real error because I removed the trailing '*', assuming that files will always end with the tag and then immediately their extension. It seems in this test that's not the case but I can't get it to run properly on my machine so I can't check the file names. Don't have time to work on this at the moment. :/

tensionhead:
Ok I see, from our side this is not top-priority, as opposed to many other open issues.. If you find time again some time in the future would be great of course :slightly_smiling_face:

dfsp-spirit:
I am closing this for now, feel free to re-open when we continue working on it.

Issue #357, opened 2022-10-06
syncopy.CrossSpectralData.singlepanelplot() returns None


dfsp-spirit:
syncopy.CrossSpectralData.singlepanelplot() returns None instead of `(fig, ax)`-tuple, like other datatypes (AnalogData, SpectralData).

This means one cannot set a plot title or axis limits, which is quite inconvenient when plotting many things in a script.

dfsp-spirit:
After inspecting the code, the fig and ax can be accessed after plotting from the data object, as `myCrossSpectralDatainstance.fig` and `myCrossSpectralDatainstance.ax`.

So this may be a non-issue.

tensionhead:
Got addressed in #365

PR #365, opened 2022-10-19
349 plotting does not support custom dimords

tensionhead:
This PR actually addreses a bunch of minor bugs, that appeared during manual testing:

- `spectralDTypes` in `shared.const_def` now contain all supported outputs
- multi-taper parameters now are aware of the maximal smoothing bw possible (edge case)
- fixed dimord for AnalogData, for other datatypes it is most likely not needed, as the frontend results always have the default dimord (how would you get a `SpectralData` with non-default dimord? ..we don't support reading FT spectral data atm anyways)
- now for `output != 'pow'` there is no general log10, so imaginary/real parts of spectra/coherence can also be plotted nicely (they of course can have negative values)
- cut back on the test helper selection generation, one random combination per toi/toilim setting is enough 
- also addressed #357, I hope there are no undesired side effects.. tests are not complaining though

Author Guidelines
-----------------

Reviewer Checklist
------------------
- [x] Code layout
  - [x] No commented-out code?

### DIFF
--- a/syncopy/plotting/mp_plotting.py
+++ b/syncopy/plotting/mp_plotting.py
@@ -253,9 +253,9 @@ def plot_CrossSpectralData(data, **show_kwargs):
     # get the data to plot
     data_y = data.show(**show_kwargs)
 
-    # create the axes and figure if needed
-    # persistent axes allows for plotting different
-    # channel combinations into the same figure
+    # Create the axes and figure if needed.
+    # Persistent axes allow for plotting different
+    # channel combinations into the same figure.
     if not hasattr(data, 'ax'):
         fig, data.ax = _plotting.mk_line_figax(xlabel, ylabel)
     _plotting.plot_lines(data.ax, data_x, data_y, label=label)
--- a/syncopy/plotting/mp_plotting.py
+++ b/syncopy/plotting/mp_plotting.py
@@ -46,7 +46,7 @@ def plot_AnalogData(data, shifted=True, **show_kwargs):
         trl = 0
 
     # get the data to plot
-    data_x = plot_helpers.parse_toi(data, trl, show_kwargs)    
+    data_x = plot_helpers.parse_toi(data, trl, show_kwargs)
     data_y = data.show(**show_kwargs)
     # 'time' and 'channel' are the only axes
     if data._defaultDimord != data.dimord:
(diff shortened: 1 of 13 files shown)


syncopy/plotting/mp_plotting.py
```diff
@@ -63,7 +66,7 @@ def plot_AnalogData(data, shifted=True, **show_kwargs):
         return
 
     elif nAx > pltConfig['mMaxAxes']:
-        SPYWarning("Please select max. {pltConfig['mMaxAxes']} channels for a multipanelplot!")
+        SPYWarning(f"Please select max. {pltConfig['mMaxAxes']} channels for a multipanelplot!")
```
dfsp-spirit, line 69: oh yes, forgetting that `f` is a common bug.

syncopy/tests/test_plotting.py
```diff
@@ -23,13 +23,15 @@ class TestAnalogPlotting():
     nSamples = 300
     adata = synth_data.AR2_network(nTrials=nTrials,
                                    AdjMat=np.zeros(nChannels),
-                                   nSamples=nSamples)
+                                   nSamples=nSamples,
+                                   seed=helpers.test_seed)
```
dfsp-spirit, line 27: Nice, it's a good idea to have one seed in a common place.

dfsp-spirit:
Thanks for this great contribution!

I have one general question that we should discuss soon maybe, that came up during the review, but is not specific to this PR: when do we give warnings, and when errors? E.g., the plotting functions just print something (technically not even a warning) without plotting anything in case of invalid input args.

Throwing an error would also make it easier to test for the exceptions, and to increase coverage.

tensionhead:
> Thanks for this great contribution!
> 
> I have one general question that we should discuss soon maybe, that came up during the review, but is not specific to this PR: when do we give warnings, and when errors? E.g., the plotting functions just print something (technically not even a warning) without plotting anything in case of invalid input args.
> 
> Throwing an error would make it easier to test for the exceptions, and to increase coverage. It may have other downsides, though.

Yes absolutely, this discussion is much needed! For the current state of the code base, I did not want to 'kill' an analysis script by raising an exception just because some plotting did not go down as expected..

PR #372, opened 2022-11-04
341 implement sequential statistics mean stddev for syncopydata

### LINKED ISSUE #230: Averaging operator and `.show()` interface
For plotting and general post-processing of analyses, having the possibility to contract a certain dimension (be it `time`, `channel`, `trial` or maybe even `freq`) would be of much use. The general idea would be to have a method attached to Syncopy data objects akin to `selectdata`, however this time new (the averaged) data will also be written on disc. As an example suppose we have a `SpectralData` object `spec` with 10 channels and say 200 trials, a typical application might look like this:
```python
av_spec = spec.average(channels=['ch1', 'ch5', 'ch8'], trials='all')
```
This would create a new `SpectralData` object `av_spec` which now only has 1 trial and a singleton `channel` axis holding the trial average over the selected 3 channels, probably best called `avg` or sth similar.

Moreover, it would be useful to also allow the `.show` method to interface this averaging directly like this:
```python
out_arr = spec.show(channels=['ch1', 'ch5', 'ch8'], trials='all', average=True)
```
In case of singleton time and taper axes (`method='mtmfft'` with `taper='hann'`) this now would return a `(nFreq,)` shaped array. Alternatively, calling `.show()` without arguments on the averaged `SpectralData` object `av_spec`, should return the exact same array. 

Additionally it might be a good messaging policy, to inform the user about the size of the returned array in memory upon calling `.show(...)`.

### LINKED ISSUE #341: Implement sequential statistics (mean, stddev) for SyncopyData
**Is your feature request related to a problem? Please elaborate.**
Users would like to compute statistics on SyncopyData objects.

**Do you have a solution in mind? Please elaborate**
We need to store these as new datasets in the hdf5, but it should be enough to do this sequentially (without dask), so we do not need a full-featured cF (and another virtual hdf5 dataset). We will just compute them piece by piece on the local machine.

tensionhead:
## Basic statistics

This PR introduces `spy.mean`, `spy.var`, `spy.std` and `spy.median`, also available as methods for all `ContinuousData` types. All statistical operation can be performed over any dimension a Syncopy data object might have plus over trials. 

For operations **not along trials** a standard CR `NumpyStatDim` was implemented, which takes all the usual keywords (`select`, `parallel`) and then recures in the associated cF to the respective NumPy functions (`np.mean(axis=...)` , ...). Those statistics hence can be computed in parallel and also work for non-equal trials.

For operations **along trials** a different path needs to be followed, as except for the averaging via `keeptrials=False,` computing statistics over trials is out of scope of the CRs. Here the functions associated with the new `trial_statistics` sequentially stream the data on a trial-by-trial basis and basically sum it into *one* ordinary `np.array` until the `.trials` iterator is exhausted. Then a new Syncopy object is created with that array as the data. Calculting things like the variance this way *manually* is generally of course less efficient than using e.g. numpy routines, BUT we keep our memory-safe promise that as long as 2 trials fit into memory everything is fine!
The only setback is that I did not find a way to achieve this for the `median` operation, as here the only manual way I could think of would be to sort every(!) dimensional entry along the trial direction separately. Something which is just so awkward and slow to be of any practical use imho. So `median` along trials is not supported atm, and I am open for suggestions here :slightly_smiling_face:  

closes #341 
closes #230 

Author Guidelines
-----------------

Reviewer Checklist
------------------
- [x] Code layout
  - [x] No commented-out code?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
 - Added additional .info entries for Granger analysis, indicating details about the computation.
 - Added additional .info entries for FoooF results, e.g. Gaussian fit parameters.
 - Fix bug #365, plotting supports custom dimords now.
+- Support starting a connectivity analysis from a SpectralData instance (#364).
 
 ### CHANGED
 - Maximal brute force regularization parameter for Granger increased to 1e-1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
 ## [Unreleased]
 
 ### NEW
+- Added basic statistics (mean, std, var and median) for Syncopy data objects
 - Added additional .info entries for Granger analysis, indicating details about the computation.
 - Added additional .info entries for FoooF results, e.g. Gaussian fit parameters.
 - Fix bug #365, plotting supports custom dimords now.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,9 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
 ## [Unreleased]
 
 ### NEW
-- Added additional .info entries for Granger analysis, indicating details about the computation
-- Added additional .info entries for FoooF results, e.i. Gaussian fit parameters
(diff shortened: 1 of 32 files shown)


dfsp-spirit (approved):
Thaks a lot, this really brings us forward.

We now have a bunch of different methods that compute the mean, but I think this is the most appropriate solution.

At one point we should add a separate page / tutorial to our documentation that illustrates how to use it, so people find it.

PR #385, opened 2022-12-02
384 deprecate toitoilim and foifoilim for selectdata

tensionhead:
## Make `selectdata` more similar to `ft_selectdata.m`

- `toi/toilim` and `foi/foilim` gone in favor of `latency` and `frequency` as in [ft_selectdata](https://github.com/fieldtrip/fieldtrip/blob/master/utilities/ft_selectdata.m)
- (only!) `toi` and `toilim` were tolerant if the upper or lower boundary were outside of the range of a trial time, leading to inconsistent or at least intransparent results  
- This is now no longer the case: `latency` either throws an error if no trial fits, or kicks out trials which are not fitting to the desired time window retaining only the fitting trials
- this means, that every `latency` selection now returns **timelocked** data , which should be pretty neat to continue with a freqanalysis and so on..
- some selection focused tests could be greatly simplified as no `toi` and `foi` selections need to be tested anymore

### Important

- `foi/foilim` and `toi/toilim` remain valid parameters for `freqanalysis`, as also in [ft_freqanalysis](https://github.com/fieldtrip/fieldtrip/blob/master/ft_freqanalysis.m)
- `foi/foilim` remain valid for `connectivityanalysys` if a mtmfft spectral estimation is to be done on the fly from an `AnalogData` input
- in none of these cases those parameters enter into a selection!

Author Guidelines
-----------------

Reviewer Checklist
------------------
- [x] Code layout
  - [x] No commented-out code?


dfsp-spirit (approved):
Thanks a lot for this change!

PR #386, opened 2022-12-03
memory-improvements-nwb

KatharineShapcott:
Author Guidelines
-----------------

Reviewer Checklist
------------------
- [ ] Code layout
  - [ ] No commented-out code?


KatharineShapcott:
The original change didn't help the speed at all and in fact made it worse. The new changes are because I realised that if I could just assign the data directly to the hdf5 file it was much much faster. For our analysis we anyway need these two channel types split. With the addition of the container as input I can "trick" syncopy into updating this as a syncopy object (as opposed to copying an 80Gb file) by doing: 

```python
angData.data.file["trialdefinition"] = angData.trialdefinition
angData.save(filename = angData.filename, overwrite=True)
```

pantaray:
Hi! Your "trick" is actually pretty nifty :) Are you done with your modifications or are you still working on it?

KatharineShapcott:
Okay I think this is done now. One further improvement would be to write the data in parallel using acme. But since a 50Gb .nwb file -> 200 Gb .spy file now only takes ~10 minutes using just 3 Gb of memory I think I'll leave it like this for now.

syncopy/io/_load_nwb.py
```diff
@@ -106,8 +108,6 @@ def load_nwb(filename, memuse=3000):
             channels = acqValue.electrodes[:].location
```
pantaray, line 107: If I understand correctly, you create separate `AnalogData` objects for each `acquisition` in the NWB file (which makes a lot of sense!). In that case I'm not sure, if the existing channel counting logic is then still valid: https://github.com/esi-neuroscience/syncopy/blob/d5a735cb6bf9c3ad07f88c85bf0b2ef2957253f7/syncopy/io/_load_nwb.py#L118

syncopy/io/_load_nwb.py
```diff
@@ -162,7 +162,19 @@ def load_nwb(filename, memuse=3000):
 
     # Print status update to inform user
     SPYInfo(msg)
-    SPYInfo("Creating AnalogData object...")
+    
+    # Check for filename
```
pantaray, line 165: This is really elegant!

syncopy/io/_load_nwb.py
```diff
@@ -202,61 +220,63 @@ def load_nwb(filename, memuse=3000):
...
-    angData.log = msg
-    if evtData is not None:
-        evtData.log = msg
-
-    return angData, evtData
+                angDset[st_samp : end_samp, :] *= gains
+            
+        # Finalize angData
+        angData.data = angDset
+        angData.channel = acqValue.electrodes[:].location.to_list()
```
pantaray, line 274: Channels again: if channel names are not defined for the current `acquisition`, then this might file, I think.

pantaray (commented):
Looks great! I just have one comment/question regarding the channel counting/assignment.

syncopy/io/_load_nwb.py
KatharineShapcott: Ah sorry, nChannels is unused now. We can remove it

syncopy/io/_load_nwb.py
KatharineShapcott: Ah yes true, we should add back in your previous check there right?

```
if channels.unique().size == 1:
    SPYWarning("No channel names found for {}".format(acqName))
else:
    angData.channel = acqValue.electrodes[:].location.to_list()
```

syncopy/io/_load_nwb.py
pantaray: Ah, of course - sorry, didn't see that. Yes I guess removing it avoids confusion.

syncopy/io/_load_nwb.py
pantaray: Yes, I think this should do it :)

syncopy/io/_load_nwb.py
KatharineShapcott: Ha yes, copy pasted from spy.save ;)

tensionhead:
Am I overlooking sth or are we really not having a test for the `nwb` reader?!

pantaray:
Yupp, I don't think we ever had one

tensionhead:
Ok well that's not good of course.. @KatharineShapcott would you be able to trim down one of your 'production scripts' such that basic read in functionalities are covered? I could then try to whip up a proper test from your minimal example 🙂

KatharineShapcott:
> Ok well that's not good of course.. @KatharineShapcott would you be able to trim down one of your 'production scripts' such that basic read in functionalities are covered? I could then try to whip up a proper test from your minimal example 🙂

I tried to send you guys data for this before, did you manage to access it? 
https://github.com/esi-neuroscience/syncopy/commit/0a7c1427cbbe7064feba8cea517fff375431f014#commitcomment-86661238

This is an example of how I'm using it to get the lfp:

```
nwb_filename = '/path/your_nwb_file.nwb'
spy_filename = nwb_filename[:-4]+'.spy'
out = spy.load_nwb(nwb_filename, memuse=memuse)
out = list(out.values())

raw_data = None
for data in out:
    if data.__class__.__name__ == "AnalogData" and data.channel[0].startswith('CH'):
        if raw_data is not None:
            raise ValueError('2nd set of raw data found: %s' %(data.filename))
        raw_data = data
```
Which would be passed into another function like this:
```
# Bandpass filter
lfp = spy.preprocessing(raw_data, filter_class='but', freq=[lowpass, highpass], filter_type='bp', order=4, chan_per_worker=1)
# Downsample 
lfp = spy.resampledata(lfp, resamplefs=downsample_freq, method='downsample', chan_per_worker=1)
lfp.info = raw_data.info

lfp.save(container=spy_filename, tag='lfp')
```

tensionhead:
Oh, ya sorry apparently there was no follow up.. ok got the data, except for that I don't see any use of the `spy_filename` variable I think I could use your snippet to write a very basic test. I'll do it right here, that we don't have to open another branch/PR

KatharineShapcott:
Looks good to me! Maybe what is missing is a test for the bug that I found previously, that the data in the nwb file and and syncopy file are not the same. I think you should be able to check this with something like this:

```
nwbio = pynwb.NWBHDF5IO(nwb_filename, "r", load_namespaces=True)
nwbfile = nwbio.read()

spy_filename = nwb_filename[:-4]+'.spy'
out = spy.load_nwb(nwb_filename, memuse=memuse, container=spy_filename )
out = list(out.values())

for data in out:
    if data.__class__.__name__ == "AnalogData":
        data_name = data.tag
        chan0nwb = nwbfile.acquisition[data_name].data[:,0]*nwbfile.acquisition[data_name].channel_conversion[0]
        chan0spy = data.data[:,0]
        # check for equality
    elif data.__class__.__name__ == "EventData": 
      ttl_vals_nwb = nwbfile.acquisition["TTL_PulseValues"].data[:,0]
      ttl_chans_nwb = nwbfile.acquisition["TTL_Channels"].data[:,0]
  
      ttl_vals_spy = data.data[:,1]
      ttl_chans_spy = data.data[:,2]
       # check for equality
 
```

tensionhead:
Looks like the `data.tag` is not working, as it is just `None` when I try to use your last snippet. There is btw no single instance in our code base known to me, where the `.tag` attribute is used, so I consider that legacy/experimental/untested. The `load_nwb` function does not apply a tag anywhere as far as I can tell,  but as I am not the original author auf `load_nwb` and have 0 experience with the pynwb library, I am really not the ideal person to provide the full missing test I am afraid.

tensionhead:
Ok as some testing is better than no testing at all, I'll merge this as it is. @KatharineShapcott thx for your contribution!

PR #391, opened 2022-12-12
387 fix spycleanup

dfsp-spirit:
Author Guidelines
-----------------

Reviewer Checklist
------------------
- [ ] Code layout
  - [ ] No commented-out code?

### DIFF
--- a/syncopy/io/utils.py
+++ b/syncopy/io/utils.py
@@ -77,11 +77,8 @@ def cleanup(older_than=24, interactive=True, **kwargs):
     """
 
     # Make sure age-cutoff is valid
-    try:
-        scalar_parser(older_than, varname="older_than", ntype="int_like",
+    scalar_parser(older_than, varname="older_than", ntype="int_like",
                       lims=[0, np.inf])
-    except Exception as exc:
-        raise exc
     older_than = int(older_than)
 
     # For clarification: show location of storage folder that is scanned here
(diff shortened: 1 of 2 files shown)


syncopy/io/utils.py
```diff
@@ -57,7 +57,7 @@ def hash_file(fname, bsize=65536):
     return hash.hexdigest()
 
 
-def cleanup(older_than=24, interactive=True, **kwargs):
```
dfsp-spirit, line 60: I have removed the kwargs here, as they are not needed and I had a typo (`olderthan` instead of `older_than`) in my first test script that was not reported because of the kwargs and caused me a bit of a headache.

tensionhead: yes, good  idea.. there's too much `**kwargs` all over the place 😁

syncopy/io/utils.py
```diff
@@ -77,11 +77,8 @@ def cleanup(older_than=24, interactive=True, **kwargs):
     """
 
     # Make sure age-cutoff is valid
-    try:
-        scalar_parser(older_than, varname="older_than", ntype="int_like",
+    scalar_parser(older_than, varname="older_than", ntype="int_like",
```
tensionhead, line 80: yes... that's how these "parsers" should be used, they have their own exception handling/messagin

syncopy/io/utils.py
```diff
@@ -119,22 +116,24 @@ def cleanup(older_than=24, interactive=True, **kwargs):
...
+                        files = glob(os.path.join(__storage__, "*_{}_*".format(sessid)))
+                        flsList.append(files)
+                        ageList.append(round(age/24))                  # age in days
+                        usrList.append(sesslog[:sesslog.find("@")])
+                        ownList.append(sesslog[:sesslog.find(":")])
+                        sizList.append(sum(os.path.getsize(file) if os.path.isfile(file) else sum(os.path.getsize(os.path.join(dirpth, fname)) \
                                        for dirpth, _, fnames in os.walk(file)
                                        for fname in fnames) for file in files))
+            except OSError as ex:
+                print(f"Unable to open {fid}: {ex}. (Maybe already deleted.)")
```
tensionhead, line 136: Looking forward to proper logging.. but for now we could use `SPYInfo` but it's essentially just a print, so let's leave it like that

tensionhead (approved):
Ok that was a bit more involved than I expected.. thx @dfsp-spirit !

PR #400, opened 2022-12-16
394 fix structdict copy

dfsp-spirit:
Author Guidelines
-----------------

Reviewer Checklist
------------------
- [ ] Code layout
  - [ ] No commented-out code?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,15 @@
 # Changelog of SyNCoPy
 All notable changes to this project will be documented in this file.
 
+## Unreleased WIP
+
+### NEW
+
+### CHANGED
+
+### Fixed
+- fix bug #394 'Copying a spy.StructDict returns a dict'.
+
 ## [2022.12]
 
 ### NEW
--- a/syncopy/shared/tools.py
+++ b/syncopy/shared/tools.py
@@ -7,7 +7,7 @@
 import numpy as np
 import inspect
 import json
-import subprocess
+import copy
 
 # Local imports
 from syncopy.shared.errors import SPYValueError, SPYWarning, SPYTypeError
@@ -47,6 +47,12 @@ class StructDict(dict):
         else:
             ppStr = "{}"
         return ppStr
+
+    def copy(self):
+        """Overwrite the .copy method of the parent 'dict' class, otherwise copy() will return a dict."""
+        obj = type(self).__new__(self.__class__)
+        obj.__dict__.update(self.__dict__)
+        return obj
 
 
 class SerializableDict(dict):
@@ -272,4 +278,4 @@ def get_defaults(obj):
     return StructDict(dct)
 
 
-    
+
--- a/syncopy/shared/tools.py
+++ b/syncopy/shared/tools.py
@@ -48,23 +48,27 @@ class StructDict(dict):
             ppStr = "{}"
         return ppStr
 
-    def copy(self):
+    def copy(self, deep=True):
         """
-        Create a shallow-copy of this StructDict instance.
+        Create a copy of this StructDict instance.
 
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,15 @@
...
 All notable changes to this project will be documented in this file.
 
+## Unreleased WIP
+
+### NEW
+
+### CHANGED
+
+### Fixed
+- fix bug #394 'Copying a spy.StructDict returns a dict'.
```
tensionhead, line 11: Nach dem release ist vor dem release, indeed :grin:

syncopy/shared/tools.py
```diff
@@ -48,6 +47,13 @@ def __str__(self):
             ppStr = "{}"
         return ppStr
 
+    def copy(self):
+        """Overwrite the .copy method of the parent 'dict' class, otherwise copy() will return a dict."""
+        obj = type(self).__new__(self.__class__)
```
tensionhead, line 52: Ok that's as clean as it gets :+1:

syncopy/tests/test_tools.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def test_structdict_from_dict(self):
+        my_dict = {'a': 0.5, 'b': 'test', 'c' : [1, 2, 3]}
+        cfg = spy.StructDict(my_dict)
+        assert type(cfg) == spy.shared.tools.StructDict
+        assert cfg.a == 0.5
+        assert cfg.b == "test"
+        assert cfg.c == [1, 2, 3]
+
+    def test_structdict_from_dict_copy(self):
```
tensionhead, line 43: Is that test really necessary? Once the `cfg` is a `StructDict`, this test just repeats the standard copy test no?!

syncopy/tests/test_welch.py
```diff
@@ -362,6 +349,10 @@ def test_welch_rejects_invalid_output(self):
                 with pytest.raises(SPYValueError, match="output"):
                     _ = spy.freqanalysis(cfg, self.adata)
 
+    def teardown_class(cls):
```
tensionhead, line 352: yes, this we could probably need also elsewhere, maybe put it inside a `helper` module?

tensionhead (commented):
Great, this will make the `StructDict` behavior more consistent.. just one question, how can we make deep copies? The default shallow copies seem quite dangerous from a user perspective?!

syncopy/tests/test_tools.py
dfsp-spirit: yes, I will remove it. I just had the impression in between that something was fishy, but it had a different reason.

dfsp-spirit: its removed

syncopy/tests/test_welch.py
dfsp-spirit: I moved it to `tests/helpers.py`

dfsp-spirit:
> 

I have implemented a `.deepcopy()` method in [9560672](https://github.com/esi-neuroscience/syncopy/commit/95606720c6ff381ffd72d9b31c0a8e9c676d56ce).

syncopy/shared/tools.py
```diff
@@ -48,6 +48,42 @@ def __str__(self):
...
+
+        Notes
+        -----
+        Consider calling the `.deepcopy()` method to get a deep copy instead, which is typically what users want.
+        """
+        obj = type(self).__new__(self.__class__)
+        obj.__dict__.update(self.__dict__)
+        return obj
+
+    def deepcopy(self):
```
tensionhead, line 69: Wondering if we should just wire `copy()` to this deepcopy method as a default?! I think from a user perspective shallow copies are hardly a useful feature, even though shallow copies are more or less Python's default?

tensionhead (approved):
Ok great, now ppl can safely throw around `.copy()` on their `cfg`'s 👍

PR #402, opened 2022-12-23
Merge Doc sprint2212

tensionhead:
- selections
- welch's method
- new theme: sphinx-book-theme


syncopy/specest/freqanalysis.py
```diff
@@ -315,8 +315,30 @@ def freqanalysis(data, method='mtmfft', output='pow',
 
     Examples
     --------
-    Coming soon...
 
+    Generate 10 seconds of white noise, sampled at 1000 Hz:
```
tensionhead, line 319: Great that we have at least one example now @dfsp-spirit :)

PR #415, opened 2023-01-10
408 check out contiguous channel property for discrete data

tensionhead:
Changes Summary
---------------------------
This PR does 2 things, but can't follow the original idea about *contiguous channel labels*
- new attributes for `SpikeData`: `.channel_idx` and `.unit_idx`, holding the `np.unique` results 
- `.channel`/`.unit` property lookups are no longer re-computed and hence as fast as they should be
- checks for integer data type of `.data` for `DiscreteData`

I hope I captured what you had in mind @kshapcott, here's an example how it works now:
```python
spd = spy.SpikeData(data = 4 * np.ones((5,3), dtype=int), channel=['my_only_channel'])

# assigned label
spd.channel
>>> array(['my_only_channel'], dtype='<U15')

# there's only channel 4
spd.channel_idx
>>> array([4])

# default label
spd.unit
>>> array(['unit05'], dtype='<U6')

# there's only unit 4
spd.unit_idx
>>> array([4])
```
Note that both `channel` and `unit` use 0-indexing now, before only channel was 0-based for some reason.

Reviewer Checklist
------------------

### DIFF
--- a/syncopy/datatype/discrete_data.py
+++ b/syncopy/datatype/discrete_data.py
@@ -356,9 +356,6 @@ class SpikeData(DiscreteData):
         # this comes from BaseData
         self._set_dataset_property(inData, "data")
 
-        # set the default channel labels
-        self.channel = self._get_default_channel()
-
     @property
     def channel(self):
         """ :class:`numpy.ndarray` : list of original channel names for each unit"""
@@ -373,8 +370,7 @@ class SpikeData(DiscreteData):
         elif self.data is None and chan is not None:
             raise SPYValueError("Cannot assign `channel` without data. " +
                                 "Please assign data first")
-        else:
-            # chan was None
+        elif chan is None:
             self._channel = chan
             return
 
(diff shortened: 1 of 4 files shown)


tensionhead:
Ok.. now the coverage complains about lines I haven't touched, so I won't grind it any further for the moment

syncopy/datatype/discrete_data.py
```diff
@@ -51,8 +51,15 @@ def data(self):
 
     @data.setter
     def data(self, inData):
+        """ Also checks for integer type of data """
+        # this comes from BaseData
         self._set_dataset_property(inData, "data")
 
+        if inData is not None:
+            # probably not the most elegant way..
+            if 'int' not in str(self.data.dtype):
```
KatharineShapcott, line 60: I think this would be the "right" way to do it 
`if np.issubdtype(self.data.dtype, np.integer)`

syncopy/datatype/discrete_data.py
```diff
@@ -324,68 +334,125 @@ class SpikeData(DiscreteData):
     _stackingDimLabel = "sample"
     _selectionKeyWords = DiscreteData._selectionKeyWords + ('channel', 'unit',)
 
+    def _compute_unique(self):
```
KatharineShapcott, line 337: `_compute_unique_idx` maybe more clear?

KatharineShapcott (changes requested):
Looks good! Should be soooo much faster for my data in combination with 403. I can make those changes if you agree.

syncopy/datatype/discrete_data.py
```diff
@@ -153,8 +160,8 @@ def trialid(self, trlid):
             return
 
         if self.data is None:
-            print("SyNCoPy core - trialid: Cannot assign `trialid` without data. " +
-                  "Please assing data first")
+            SPYError("SyNCoPy core - trialid: Cannot assign `trialid` without data. " +
+                     "Please assing data first")
```
KatharineShapcott, line 164: typo maintained ;)

syncopy/datatype/discrete_data.py
tensionhead: uuh, yes that looks much better!

syncopy/datatype/discrete_data.py
tensionhead: Fine with me :)

syncopy/datatype/discrete_data.py
tensionhead: The "cannot" :see_no_evil:  ?!

KatharineShapcott: "Please assing data first" 😁

tensionhead:
> Looks good! Should be soooo much faster for my data in combination with 403. I can make those changes if you agree.

Thx, yes please go ahead with the little changes you proposed :+1:

tensionhead:
> Looks good! Should be soooo much faster for my data in combination with 403. I can make those changes if you agree.

Thx, yes please go ahead with the little changes you proposed :+1:

tensionhead:
Ok, thx for the review and the final fix ups! Shall I merge it with my super powers?

KatharineShapcott (approved):
Approved

tensionhead:
> Wait but aren't we still failing the code coverage check?

yes, but this is not entirely reliable sadly.. we merged things in that state before ;)

PR #422, opened 2023-01-16
NEW: TrialIndexer

tensionhead:
Changes Summary
----------------
- replaces the somewhat over-engineered yet still inefficient `Indexer`, backing the `.trials` property
- the new `TrialIndexer` can also be used both as an iterable or via explicit (single element) indexing

Reviewer Checklist
------------------

### DIFF
--- a/syncopy/__init__.py
+++ b/syncopy/__init__.py
@@ -127,7 +127,7 @@ from .plotting import *
 from .preproc import *
 
 # Register session
-__session__ = datatype.base_data.SessionLogger()
+__session__ = datatype.util.SessionLogger()
 
 # Override default traceback (differentiate b/w Jupyter/iPython and regular Python)
 from .shared.errors import SPYExceptionHandler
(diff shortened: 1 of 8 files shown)


syncopy/tests/test_basedata.py
```diff
@@ -161,6 +161,44 @@ def test_trialdef(self):
...
+            data.selection.trials[1]
+
+        # check that invalid trial indexing gets catched
+        with pytest.raises(SPYTypeError, match='trial index'):
+            data.trials[range(4)]
+        with pytest.raises(SPYTypeError, match='trial index'):
+            data.trials[2:3]
+        with pytest.raises(SPYTypeError, match='trial index'):
+            data.trials[np.arange(3)]
+        return data
```
KatharineShapcott: Should the test actually be returning `data`?

tensionhead: Of course not.. I remove that already with the last commit :smile:

KatharineShapcott (approved):
I think it all looks good!

PR #427, opened 2023-01-27
CHG: Avoid list comprehension for discrete select

KatharineShapcott:
Changes Summary
----------------
- Discrete data selection speedup

Reviewer Checklist
------------------


KatharineShapcott:
Yeah I think that was the biggest, but for looping through trialArr multiple times per trial was also not good when it can include 50,000 spikes per trial!

PR #409, opened 2023-01-04
Logging

dfsp-spirit:
This implements logging. Logging uses the standard Python `logging` module, which is quite flexible.

Similar to SpyError/SpyParallelError, there are now 2 loggers (accessible via `syncopy.shared.log.get_logger()` and `...get_par_logger()` respectively) for local and remote/parallel logging. Both log to files by default, the local one also logs to console. Exceptions are also logged. To manually log something, the old logging functions (like SPYInfo, SPYWarning) can be used, and a new all-in-one function SPYlog was added.

Note that the default loglevel is IMPORTANT, so you will NOT see any SPYInfo messages unless you change that. (The new documentation on logging explains how to do that.)

Other changes:

* allow silencing of syncopy on-import message (via dotfile or env var)
* disable tqdm output in logs (but keep it in interactive mode)
* add debug logging messages in various parts of the source code. This is a WIP and should be continued in the future when adding new code.

Author Guidelines
-----------------
- [x] Is the change set **< 600 lines**?
- [ ] Was the code checked for memory leaks/performance bottlenecks?
- [ ] Is the code running locally **and** on the ESI cluster?
- [ ] Is the code running on all supported platforms?

Reviewer Checklist
------------------
- [ ] Are testing routines present?
- [ ] Do **parallel** loops have a set length and correct termination conditions?
- [ ] Do objects in the global package namespace perform proper parsing of their input? 
- [ ] Do code-blocks provide novel functionality, i.e., no re-factoring using builtin/external packages possible?
- [ ] Code layout
  - [ ] Is the code PEP8 compliant?
  - [ ] Does the code adhere to agreed-upon naming conventions?
  - [ ] Are keywords clearly named and easy to understand?
  - [ ] No commented-out code?
- [ ] Are all docstrings complete and accurate?
- [ ] Is the CHANGELOG.md up to date?

PR #428, opened 2023-01-30
FIX: load_nwb creates integer EventData

KatharineShapcott:
Changes Summary
----------------
- Use int datatype for loaded EventData

Reviewer Checklist
------------------

### DIFF
--- a/syncopy/io/load_nwb.py
+++ b/syncopy/io/load_nwb.py
@@ -197,7 +197,7 @@ def load_nwb(filename, memuse=3000, container=None):
 
         evtData = EventData(dimord=["sample","eventid","chans"], filename=filename)
         h5evt = h5py.File(evtData.filename, mode="w")
-        evtDset = h5evt.create_dataset("data", dtype=np.result_type(*ttlDtypes),
+        evtDset = h5evt.create_dataset("data", dtype=int,
                                        shape=(ttlVals[0].data.size, 3))
         # Column 1: sample indices
         # Column 2: TTL pulse values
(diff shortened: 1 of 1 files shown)


syncopy/io/load_nwb.py
```diff
@@ -197,7 +197,7 @@ def load_nwb(filename, memuse=3000, container=None):
 
         evtData = EventData(dimord=["sample","eventid","chans"], filename=filename)
         h5evt = h5py.File(evtData.filename, mode="w")
-        evtDset = h5evt.create_dataset("data", dtype=np.result_type(*ttlDtypes),
+        evtDset = h5evt.create_dataset("data", dtype=int,
```
tensionhead, line 200: Okay that looks much simpler of course.. we can be sure only integers are to be expected here?

KatharineShapcott: No they are floats but they need to be converted otherwise it crashes.

tensionhead: Ok.. so does that mean there could be events encoded with floats? I think we only enforce integere types for `SpikeData` now, not for `EventData`?!

KatharineShapcott: Events should always be integers but they're not stored like that in NWB. No it's enforced for both, this crashes now.

tensionhead: Uh, but apparently rightfully so :slightly_smiling_face: Ok then let's do it like you proposed!

KatharineShapcott: We could print a warning that we're casting?

tensionhead: Maybe at the debug level, we have proper logging now with #409 But as you said, the nwb reader should anyways be considered experimental :sweat_smile:

PR #430, opened 2023-01-30
Add log function to syncopy root namespace

dfsp-spirit:
Changes Summary
----------------
- Adds a `syncopy.log()` function that can be used to log without first importing stuff from submodules, and supports both the sequential and the parallel loggers:

```python
import syncopy as spy
spy.log("We are about to get real.", level="IMPORTANT")
spy.log("oh no, an error in parallel code!", level="ERROR", par=True)
```

Reviewer Checklist
------------------

### DIFF
--- a/syncopy/shared/errors.py
+++ b/syncopy/shared/errors.py
@@ -369,7 +369,7 @@ def SPYLog(msg, loglevel="INFO", caller=None):
     logfunc(PrintMsg.format(caller=" <" + caller + ">" if len(caller) else caller,
                           msg=msg))
 
-def log(msg, level="INFO", par=False, caller=None):
+def log(msg, level="IMPORTANT", par=False, caller=None):
     """
     Log a message using the Syncopy logging setup.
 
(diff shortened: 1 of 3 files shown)


syncopy/shared/errors.py
```diff
@@ -361,6 +369,30 @@ def SPYLog(msg, loglevel="INFO", caller=None):
     logfunc(PrintMsg.format(caller=" <" + caller + ">" if len(caller) else caller,
                           msg=msg))
 
+def log(msg, level="INFO", par=False, caller=None):
```
tensionhead, line 372: So with the default log-level (`"IMPORTANT") this will emit nothing when just being called with: `spy.log("my message")?

dfsp-spirit: yes. I have changed the default log level of the function to IMPORTANT, so we need to do less typing.

tensionhead (approved):
Great, with this `spy.log()` can be conveniently used to drop persistent messages!

PR #433, opened 2023-02-01
Spike waveform

KatharineShapcott:
Changes Summary
----------------
- New waveform property for SpikeData objects

Reviewer Checklist
------------------

### DIFF
--- a/syncopy/datatype/base_data.py
+++ b/syncopy/datatype/base_data.py
@@ -812,7 +812,9 @@ class BaseData(ABC):
         """
         for propertyName in self._hdfFileDatasetProperties:
             dsetProp = getattr(self, "_" + propertyName)
+            print(f"_get_backing_hdf5_file_handle: checking prop '{propertyName}'")
             if isinstance(dsetProp, h5py.Dataset):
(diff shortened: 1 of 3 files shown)


KatharineShapcott:
@dfsp-spirit Can you take a look at this and see what I'm doing wrong? Maybe adding in the chunking was a mistake?

dfsp-spirit:
Yes, I will have a look!

dfsp-spirit:
@KatharineShapcott  I think I have an idea what is going on, I will try to fix it.

dfsp-spirit:
@KatharineShapcott Something is really off with this branch, I noticed it did not have the latest dev also. I have transferred your changes to this new PR and branch `waveform`: [PR 437](https://github.com/esi-neuroscience/syncopy/pull/437) that is now green.

I also added some tests for the current functionality.

What you should think about is the behavior you want for selections, e.g., what should happen to the waveform if a selection is applied to the data (I guess you want to only select the respective part of the waveform then). You should add a test and selection logic for that, if such a thing is needed. If not, I guess this is pretty much it.

dfsp-spirit:
Closing in favor of PR 437.

Issue #238, opened 2022-03-16
SpikeData "waveform" property


KatharineShapcott:
Fieldtrip's spike structure has an optional "waveform" field. It would be useful to have a similar "waveform" property in the SpikeData class so that the type of spikes (e.g. inhibitory or excitatory) can be identified. 

Quote from https://www.fieldtriptoolbox.org/tutorial/spike/

"The first dimension of spike.waveform{i} is ‘leads’. For tetrode recordings, multiple leads per electrode are available, in which case the first dimension of spike.waveform{i} would have been of size 4. 
The second dimension of waveform contains the samples. In this case one sample corresponds to 1/40000 seconds. 
The third dimension of spike.waveform{i} equals the length of spike.timestamp{i}, such that a waveform is present for every spike (‘spike’ dimension)."

We would only need this for sorted spikes though, which won't happen for a few more months.

tensionhead:
Ah ok I see, thx @KatharineShapcott for bringing this up! @joschaschmiedt is that something you could take care of?

KatharineShapcott:
Seems like other well used formats are also three dimensional. 

> Now we can retrieve waveforms per unit on-the-fly. The waveforms shape is (num_spikes, num_sample, num_channel):

https://spikeinterface.readthedocs.io/en/latest/modules/core/plot_4_waveform_extractor.html

In this case the extra dimension is channel which I think is definitely more general than `lead`

tensionhead:
mmh.. what is `num_spikes` and `num_sample`?

KatharineShapcott:
`num_spikes` is the total number of spikes, `num_sample` is the number of samples that makes up the waveform

KatharineShapcott:
They have a seperate storage for each unit (or channel in our case) though. This might makes sense since the most common operation is to mean or median across the unit and plot the average waveform. But it will mess up all our indexing

dfsp-spirit:
This is implemented in [PR437](https://github.com/esi-neuroscience/syncopy/pull/437).

tensionhead:
Done with #437

PR #437, opened 2023-02-09
Waveform

dfsp-spirit:
Changes Summary
----------------
- Introduces a new dataset named `waveform` for `SpikeData` instances, see #238
- Selecting data on the main dataset automatically selects the waveforms/raw data for the appropriate spikes. (Makes no sense for in-place selections).
- Can save and load SpikeData including a wavefrom dataset.

Limitations:
- We expect the waveform dataset to have same size as the main `data` dataset for dimensions 1 to 2. It can have a 3rd dim.

Reviewer Checklist
------------------

### DIFF
--- a/syncopy/datatype/base_data.py
+++ b/syncopy/datatype/base_data.py
@@ -421,13 +421,13 @@ class BaseData(ABC):
             act = "backing HDF5 file is closed"
             raise SPYValueError(legal=lgl, actual=act, varname="data")
 
-        # Ensure dataset has right no. of dimensions
-        if inData.ndim != ndim:
-            lgl = "{}-dimensional data".format(ndim)
-            act = "{}-dimensional HDF5 dataset".format(inData.ndim)
-            raise SPYValueError(legal=lgl, varname="data", actual=act)
-
         if propertyName == "data":
+            # Ensure dataset has right no. of dimensions
+            if inData.ndim != ndim:
+                lgl = "{}-dimensional data".format(ndim)
+                act = "{}-dimensional HDF5 dataset".format(inData.ndim)
+                raise SPYValueError(legal=lgl, varname="data", actual=act)
+
             self._check_dataset_property_discretedata(inData)
             self.filename = inData.file.filename
         else:
--- a/syncopy/datatype/base_data.py
+++ b/syncopy/datatype/base_data.py
@@ -2135,15 +2135,6 @@ class Selector:
             # Finally, prepare new `trialdefinition` array
             self.trialdefinition = data
 
-            # Ensure that `self.waveform` gets selected correctly
-            if self._dataClass == "SpikeData":
-                if data.waveform is not None:
-                    waveform = []
-                    for tk, trialno in enumerate(self.trial_ids):
-                        sel = getattr(self, "_{}".format(actualSelections[0]))[tk]
-                        waveform.append(data.waveform[data._trialslice[trialno],:][sel,:])
-                    self.waveform = np.vstack(waveform)
-
             return
 
         # Count how many lists we got
--- a/syncopy/datatype/base_data.py
+++ b/syncopy/datatype/base_data.py
@@ -371,15 +371,15 @@ class BaseData(ABC):
             if self.mode == "r":
                 lgl = "dataset with write or copy-on-write access"
                 act = "read-only file"
-                raise SPYValueError(legal=lgl, varname="mode", actual=act)
+                raise SPYValueError(legal=lgl, varname=propertyName, actual=act)
             if prop.shape != inData.shape:
                 lgl = "dataset with shape {}".format(str(prop.shape))
                 act = "data with shape {}".format(str(inData.shape))
-                raise SPYValueError(legal=lgl, varname="data", actual=act)
+                raise SPYValueError(legal=lgl, varname=propertyName, actual=act)
             if prop.dtype != inData.dtype:
                 lgl = "dataset of type {}".format(prop.dtype.name)
                 act = "data of type {}".format(inData.dtype.name)
-                raise SPYValueError(legal=lgl, varname="data", actual=act)
+                raise SPYValueError(legal=lgl, varname=propertyName, actual=act)
             prop[...] = inData
 
         # or create backing file on disk
--- a/syncopy/datatype/base_data.py
+++ b/syncopy/datatype/base_data.py
@@ -207,7 +207,7 @@ class BaseData(ABC):
 
         supportedSetters[type(inData)](inData, propertyName, ndim=ndim)
 
-    def _unregister_dataset(self, propertyName, del_from_file=True):
+    def _unregister_dataset(self, propertyName, del_from_file=True, del_attr=True):
         """
         Unregister and delete an additional dataset from the Syncopy data object,
         and optionally delete it from the backing hdf5 file.
@@ -220,14 +220,17 @@ class BaseData(ABC):
                 The name of the entry in `self._hdfFileDatasetProperties` to remove.
                 The attribute named `'_' + propertyName` of this SyncopyData object will be deleted.
             del_from_file: bool
-                Whether to also remove the dataset named 'propertyName' from the backing hdf5 file on disk.
+                Whether to remove the dataset named 'propertyName' from the backing hdf5 file on disk.
+            del_attr: bool
+                Whether to remove the dataset attribute from the Syncopy data object.
         """
-        if propertyName in self._hdfFileDatasetProperties:
-            tmp_list = list(self._hdfFileDatasetProperties)
-            tmp_list.remove(propertyName)
-            self._hdfFileDatasetProperties = tuple(tmp_list)
-        if hasattr(self, "_" + propertyName):
-            delattr(self, "_" + propertyName)
+        if del_attr:
+            if propertyName in self._hdfFileDatasetProperties:
+                tmp_list = list(self._hdfFileDatasetProperties)
+                tmp_list.remove(propertyName)
+                self._hdfFileDatasetProperties = tuple(tmp_list)
+            if hasattr(self, "_" + propertyName):
+                delattr(self, "_" + propertyName)
         if del_from_file:
             if self.mode == "r":
                 lgl = "HDF5 dataset with write or copy-on-write access"
--- a/syncopy/datatype/base_data.py
+++ b/syncopy/datatype/base_data.py
@@ -2106,7 +2106,7 @@ class Selector:
                     chanPerTrial.append(rawChanInTrial[combinedSelect])
                 elif areShuffled:
                     combinedSelect = combinedSelect.tolist()
-                    
+
                 # The usual list -> slice conversion (if possible)
                 if len(combinedSelect) > 1:
                     selSteps = np.diff(combinedSelect)
@@ -2134,6 +2134,15 @@ class Selector:
 
             # Finally, prepare new `trialdefinition` array
             self.trialdefinition = data
+
+            # Ensure that `self.waveform` gets selected correctly
+            if self._dataClass == "SpikeData":
+                if data.waveform is not None:
+                    waveform = []
+                    for tk, trialno in enumerate(self.trial_ids):
+                        sel = getattr(self, "_{}".format(actualSelections[0]))[tk]
+                        waveform.append(data.waveform[data._trialslice[trialno],:][sel,:])
+                    self.waveform = np.vstack(waveform)
 
             return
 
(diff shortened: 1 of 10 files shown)


KatharineShapcott:
Hi Tim,
I tried to add the selection logic in this commit: 63f6b538ad0d34c637e32b78719bb1a2cf54e698
Is there a better way to do it?
Best,
Katharine

dfsp-spirit:
The selection logic in 63f6b538 is not going to work. Sadly it is not that easy:

* Selection creates a new file on disk, and copies (a subset of) the main dataset (e.g., `myspikedata.data`) from the original HDF5 file there. There currently is no handling whatsoever for extra datasets, so the output object will not even you a `.waveform` dataset after a selection.
* Only the main dataset `.data` is know to the parallelization machinery, and the PRs do not know anything about other datasets. We also have some cases where the shape of the additional datasets is different from the main data shape, so we cannot handle them automatically at all.

This means that getting the selection from the main dataset `.data` to apply to the extra dataset `.waveform` will require a lot of work. I will talk to Gregor about it in a meeting this afternoon.

KatharineShapcott:
I don't think anyone needs to do any parallel operations on `waveform` but I guess they would expect it to be selected by `selectdata`. 
I was attempting to create the new waveform dataset inside the `_make_consistent` function because that's always run at the end of `selectdata` so has information about all selections that will take place and it is only for discrete data types. Can we somehow make only the `waveform` persist? Something around here?

https://github.com/esi-neuroscience/syncopy/blob/e1e28e622a62cec55f22f59c478255733ce81f3e/syncopy/datatype/methods/selectdata.py#L315-L317

And then add:

```
    out.waveform = data.selection.waveform
```

KatharineShapcott:
Or actually that would need to be after `.data` exists right? Hmmm okay not so easy indeed....

dfsp-spirit:
We discussed this on Friday and I am looking into how we can get the selections to work.

dfsp-spirit:
Hi @KatharineShapcott, this is ready for review from my point of view.

If you find the time, feel free to play around with this branch and see if it behaves like you expect.

My interpretation on how ppl would use it can be seen [here in the tests](https://github.com/esi-neuroscience/syncopy/blob/waveform/syncopy/tests/test_discretedata.py#L454).

syncopy/datatype/base_data.py
```diff
@@ -207,7 +207,7 @@ def _register_dataset(self, propertyName, inData=None):
 
         supportedSetters[type(inData)](inData, propertyName, ndim=ndim)
 
-    def _unregister_dataset(self, propertyName, del_from_file=True):
+    def _unregister_dataset(self, propertyName, del_from_file=True, del_attr=True):
```
tensionhead, line 210: yup, makes sense here! Real dynamic properties would be the most flexible solution here, let's see maybe we still manage that..

syncopy/datatype/discrete_data.py
```diff
@@ -506,6 +506,35 @@ def _get_unit(self, trials, units=None):
...
+    @waveform.setter
+    def waveform(self, waveform):
+        """Set a waveform dataset from a numpy array, `None`, or an `h5py.Dataset` instance."""
+        if self.data is None:
+            if waveform is not None:
+                raise SPYValueError(f"non-empty SpikeData", "cannot assign `waveform` without data. " +
+                                    "Please assign data first")
+        if waveform is None:
+            self._set_dataset_property(waveform, 'waveform') # None
+            self._unregister_dataset("waveform", del_attr=False)
```
tensionhead, line 530: Nice, this register/unregister approach really works nicely!

syncopy/datatype/discrete_data.py
```diff
@@ -506,6 +506,35 @@ def _get_unit(self, trials, units=None):
...
+
+    @waveform.setter
+    def waveform(self, waveform):
+        """Set a waveform dataset from a numpy array, `None`, or an `h5py.Dataset` instance."""
+        if self.data is None:
+            if waveform is not None:
+                raise SPYValueError(f"non-empty SpikeData", "cannot assign `waveform` without data. " +
+                                    "Please assign data first")
+        if waveform is None:
+            self._set_dataset_property(waveform, 'waveform') # None
```
tensionhead, line 529: It's a bit funny, with `.data` having shape `(1000, 3)` and trying to set a waveform with shape `(1000, 2)` this gets catched here with:
```python
>>> SPYValueError: Invalid value of `data`: 'data with shape (1000, 2, 10)'; expected dataset with shape (1000, 3, 10)
```
So probably the exception should say "Invalid value of `waveform` here"?! 
When trying with a waveform with shape `(100, 3)` I get:
```python
>>> SPYValueError: Invalid value of `waveform shape[0]=100 must equal nSpikes=1000. Please create one waveform per spike in data.`: expected wrong size waveform
```
Meaning that this dataset setter does not catch it, and hence you need the 2nd check below?

syncopy/datatype/discrete_data.py
```diff
@@ -560,6 +589,10 @@ def __init__(self,
                          samplerate=samplerate,
                          dimord=dimord)
 
+        self._hdfFileDatasetProperties +=  ("waveform",)
```
tensionhead, line 592: Would that not be better placed into `SpikeData` instead of the upstream general `DiscreteData` class?

syncopy/datatype/methods/selectdata.py
```diff
@@ -363,6 +365,23 @@ def selectdata(data,
     selectMethod.compute(data, out, parallel=kwargs.get("parallel"),
                          log_dict=log_dct)
 
+    # Handle selection of waveform for SpikeData objects
+    if type(data) == spy.SpikeData and data.waveform is not None:
+        if inplace:
+            spy.log("Inplace selection of SpikeData with waveform not supported for the waveform.", level="INFO")
```
tensionhead, line 371: So the only real use case for inplace selections from the user side is for frontend runs. I could imagine some `process_metadata` hook which still takes care of the waveform after the computations then? But I am just wildly guessing here, I don't think we would absolutely need this, but maybe escalate to `Warning` level in the log?!

syncopy/datatype/methods/selectdata.py
```diff
@@ -363,6 +365,23 @@ def selectdata(data,
...
+        else:
+            fauxTrials = [data._preview_trial(trlno) for trlno in data.selection.trial_ids]
+            spikes_by_trial = [f.idx[0] for f in fauxTrials]
+            spike_idx = np.concatenate([np.array(x).ravel() for x in spikes_by_trial])
```
tensionhead, line 375: For wild sub-selections of channels/unit this will generate inhomogenuous index arrays, which as we know now are very very slow with `h5py`. However this should be also true for ordinary `SpikeData` selections, so I don't think we can do better here. Except for selecting whole trials, that's what we have those `_trialslice`'s now for, What's your take on this @kshapcott ?

syncopy/datatype/methods/selectdata.py
```diff
@@ -363,6 +365,23 @@ def selectdata(data,
...
+
+            # Copy the proper subset of the waveform dataset to `out`, the new `SpikeData` object.
+            hdf5_file_in = data._get_backing_hdf5_file_handle()
+            hdf5_file_out = out._get_backing_hdf5_file_handle()
+
+            selected_data = hdf5_file_in['/waveform'][spike_idx, :, :]
```
tensionhead, line 381: This goes all into RAM right?!

syncopy/io/load_spy_container.py
```diff
@@ -297,7 +297,16 @@ def _load(filename, checksum, mode, out):
...
         targetProperty = datasetProperty if datasetProperty == "data" else "_" + datasetProperty
-        setattr(out, targetProperty, h5py.File(hdfFile, mode="r")[datasetProperty])
+        try:
+            setattr(out, targetProperty, h5py.File(hdfFile, mode="r")[datasetProperty])
+        except KeyError:
+            if datasetProperty == "data":
+                raise SPYError("Data file {file} does not contain a dataset named 'data'.".format(file=hdfFile))
+            else:
+                spy.log(f"Dataset '{datasetProperty}' not present in HDF5 file, cannot load it. Setting to None.", level="DEBUG")
+                # It is fine if an extra dataset is not present in the file, e.g., the SpikeData waveform dataset is not present when set to None.
```
tensionhead, line 307: Yup nice, absolutely :+1:

syncopy/tests/test_discretedata.py
```diff
@@ -453,8 +451,100 @@ def test_ed_trialsetting(self):
...
+        spiked.waveform = np.ones((2, 3), dtype=int)
+        assert spiked.waveform.shape == (2, 3,)
+        spiked.waveform = None
+        assert spiked.waveform is None
+        # try to set again
+        spiked.waveform = np.ones((2, 3), dtype=int)
+        assert spiked.waveform.shape == (2, 3,)
+
+
+    def test_waveform_selection_trial(self):
```
tensionhead, line 495: Strictly speaking, also more involved channel/unit selections could be tested here.. however as this is all nicely separated and downstream of the standard selection machinery (using the same `_preview_trial.idx` and so on) it's not needed. I also just did it in an interpreter and works as expected!

tensionhead (commented):
Really great job @dfsp-spirit Reaping the benefits of our new internal dataset register/unregister machinery, and wiring it to the existing selection machinery! This strategy here might well be useful also for other additional datasets.

I have mostly comments, and maybe 2 minor suggestions below.

syncopy/datatype/discrete_data.py
dfsp-spirit: Yes, that case is handled further down the line by existing machinery. I made the exception message more specific in be9082d.

syncopy/datatype/discrete_data.py
```diff
@@ -506,6 +506,35 @@ def _get_unit(self, trials, units=None):
...
+            if waveform is not None:
+                raise SPYValueError(f"non-empty SpikeData", "cannot assign `waveform` without data. " +
+                                    "Please assign data first")
+        if waveform is None:
+            self._set_dataset_property(waveform, 'waveform') # None
+            self._unregister_dataset("waveform", del_attr=False)
+            return
+
+        if waveform.shape[0] != self.data.shape[0]:
+            raise SPYValueError(f"wrong size waveform", f"waveform shape[0]={waveform.shape[0]} must equal nSpikes={self.data.shape[0]}. " +
```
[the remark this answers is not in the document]
dfsp-spirit: fixed in3bd4797

syncopy/tests/test_discretedata.py
dfsp-spirit: yes, I also considered that, but I figured it would be kind of a duplicate test, because of the usage of `_preview_trial`. One thing that could be wrong is the index used as a parameter to `_preview_trial`, of course.

syncopy/datatype/discrete_data.py
dfsp-spirit: It _is_ in SpikeData.

tensionhead: uh, yeah you're right.. sorry!

syncopy/datatype/methods/selectdata.py
dfsp-spirit: True, fixed in f8aebec

syncopy/datatype/methods/selectdata.py
dfsp-spirit: it should be possible to add such a hook, but I would suggest not to worry about that unless we really need it (and get complaints from Kathreen or other users). I have increased the log level though, so ppl are aware of what is going on.

syncopy/datatype/methods/selectdata.py
```diff
@@ -363,6 +365,31 @@ def selectdata(data,
...
+            hdf5_file_in = data._get_backing_hdf5_file_handle()
+            hdf5_file_out = out._get_backing_hdf5_file_handle()
+
+            # Copy the waveform dataset into the new file, trial by trial to prevent memory issues.
+            ds = hdf5_file_out.create_dataset('waveform', shape=(len(spike_idx), *data.waveform.shape[1:]), dtype=data.waveform.dtype)
+            cur_new_idx = 0
+            for tidx, old_trial_indices in enumerate(spikes_by_trial):
+                num_spikes_this_trial = len(old_trial_indices)
+                #new_indices = range(cur_new_idx, cur_new_idx + num_spikes_this_trial)
+                new_indices = np.arange(cur_new_idx, cur_new_idx + num_spikes_this_trial)
```
tensionhead, line 387: Here a `slice` could be good, not sure if h5py is so clever..

tensionhead: ok just did an ad hoc `%timeit`, converting to slice would be faster, I will ad this with a commit

tensionhead (approved):
Great, thx again @dfsp-spirit! I will just make this one conversion to `slice` then ready to merge

PR #440, opened 2023-02-13
438 check workers available causes execution to hang

tensionhead:
Changes Summary
----------------
- branch off if `acme` is around earlier, and let it handle all parallel issues
- improved own dask interface: wait only for a **single worker** 

Reviewer Checklist
------------------

### DIFF
--- a/syncopy/shared/dask_helpers.py
+++ b/syncopy/shared/dask_helpers.py
@@ -32,23 +32,14 @@ def check_slurm_available():
     return has_slurm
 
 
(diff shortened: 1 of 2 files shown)


dfsp-spirit (approved):
Thanks for this work. As discussed, it seems that the way Dask works and with our independent trials, there really is no point in waiting for more workers to show up instead of simply starting to work right away.

Of course, we cannot add tests that will check this behavior on the Github CI.

PR #441, opened 2023-02-21
404 jackknifing frontend

tensionhead:
Changes Summary
----------------
- wired the jackknifing machinery into the `connectivityanalysis` frontend
- new key word `jackknife=True/False` can be used to control jackknifing
- added test for jackknifing a Granger causality analysis

Reviewer Checklist
------------------


syncopy/tests/test_statistics.py
```diff
@@ -502,6 +503,116 @@ def test_jk_coh(self, **kwargs):
         # we could still expect to get a 'significant' coherence max 5% of the time
         assert np.sum(pvals < 0.05) / coh.freq.size < 0.05
 
+        # finally fire up the frontend and compare results
+        res = spy.connectivityanalysis(adata, method='coh', jackknife=True, output=output)
+
+        assert np.allclose(res.jack_var, variance.data)
+        assert np.allclose(res.jack_bias, bias.data)
+
+    def test_jk_frontend(self):
```
dfsp-spirit, line 512: This test triggers the following warning:

    15:51:51 - WARNING: Syncopy <cross_spectra> WARNING: Multi-channel Granger analysis can be numerically unstable, it is recommended to have at least 10 times the number of trials compared to the number of channels. Try calculating in sub-groups of fewer channels!

But I guess that is fine with us, right? I usually try to make my testsuite run without warnings (which may include suppressing those I know about and which I find acceptable), but that is just my personal preference, and so far we do not care in Syncopy.

dfsp-spirit (approved):

Issue #454, opened 2023-03-24
Selecting channels in a large dataset with many trials takes ages


dfsp-spirit:
**Describe the bug**
Selecting channels in a large dataset with many trials takes ages. This was discovered by @kajal5888 

**To Reproduce**

```python
import syncopy as spy
import numpy as np
dt = spy.load("~/DataCombined_StimulusCPAll_vertical_Epoched_WindowLength50msSlidWinNr1.spy")   # file from DJ, available on ESI cs
dt2 = dt.selectdata(channel=np.arange(90, 200))  # Takes several hours, has not finished yet
```

DJ had this running for *more than 10 hours* without it finishing.

info on dt:
```python
dt
Syncopy AnalogData object with fields

                cfg : dictionary with keys ''
            channel : [220] element <class 'numpy.ndarray'>
          container : DataCombined_StimulusCPAll_vertical_Epoched_WindowLength50msSlidWinNr1.spy
               data : 182603 trials of length 50.0 defined on [9312753 x 220] float32 Dataset of size 7.63 GB
             dimord : time by channel
           filename : /home/schaefert/DataCombined_StimulusCPAll_vertical_Epoched_WindowLength50msSlidWinNr1.spy/DataCombined_StimulusCPAll_vertical_Epoched_WindowLength50msSlidWinNr1.analog
               info : dictionary with keys ''
               mode : r
         sampleinfo : [182603 x 2] element <class 'numpy.ndarray'>
         samplerate : 1017.2526245117188
          selection : None
                tag : None
               time : 182603 element list
          trialinfo : [182603 x 0] element <class 'numpy.ndarray'>
     trialintervals : [182603 x 2] element <class 'numpy.ndarray'>
             trials : 182603 element iterable

Use `.log` to see object history
```

**Note the number of trials.**

**Expected behavior**
It should perform the selection in a reasonable amount of time (seconds, not hours).

dfsp-spirit:
The issue seems to be that in the `trialdefinition` setter, `data._get_time` gets called and calls `best_match` once per trial. The `best_match` function however takes about 0.5 secs *per trial* on this dataset, resulting in a runtime of  ~ 25 hours.

dfsp-spirit:
This is being investigated in [454_chan_sel](https://github.com/esi-neuroscience/syncopy/tree/454_chan_sel).

tensionhead:
With #455 we got an performance increase of at around 2-3 orders of magnitude :rocket:

PR #455, opened 2023-03-24
454 chan sel

dfsp-spirit:
Changes Summary
----------------
See #454. This is an investigation branch that currently mostly adds logging to identify the source of the slowdown. It identified best_match as the main source of the excessive runtime, which takes 0.5 sec per trial, at 180k trials, for this dataset, leading to ~ 25h runtime.

EDIT: `best_match` in the end wasn't the culprit, an ineffecient handling of the `.time` property was. For a lean channel selection, the costly `best_match` does not need to be invoked at all..

**We were able to avoid and speed up various internal function calls, and now the runtime is down from 25h to <10 minutes.**

Reviewer Checklist
------------------


KatharineShapcott:
Apologies for the unsolicited advice, but couldn't you just temporarily remove the trial definition for a channel only selection and add it back in afterwards?

tensionhead:
Yes we also thought along those lines! Yet I think it is still interesting to find out, why a mere channel selection requires that much time/computations..

KatharineShapcott:
I suspect it might also be due to the way syncopy HDF5 files are stored, without chunking in the channel dimension I think you actually have to load in the whole file to access just one channel.

dfsp-spirit:
The selection, as everything in Syncopy, happens via trials, so we have to enumerate all trials. This branch shows that the major part of the time required to do that goes into the `best_match()` function. I will, for now, leave it at that and close this. As Katherine suggested, the workaround is to select channels before applying a trial definition like that.

tensionhead (approved):
Great, together with #489 we made another performance leap quite noticable for people with >1 000 trials, and critical for people with >10 000 trials.

Compared to current `master`, selecting 3 channels and 1000 trials from a 40 channel 10 000 trial dataset now takes 0.5 seconds instead of 55 seconds.

syncopy/datatype/selector.py
```diff
@@ -894,6 +883,7 @@ def _make_consistent(self, data):
 
                 # Update selector properties
                 for selection in actualSelections:
+                    print(getattr(self, "_{}".format(selection)), self._dataClass)
```
dfsp-spirit, line 886: there are commits missing, we have already removed this

PR #459, opened 2023-03-31
CHG: Normalize spectra independent of padding

tensionhead:
Changes Summary
----------------

- this implements Pascal's original power normalization intention: spectral power is independent of signal length/padding
- added `ft_compat` parameter for `freqanalysis --> mtmfft` to get FT behavior if needed
- no changes needed for `mtmconvol` as here the individual segments get not padded (modulo the edges)

Reviewer Checklist
------------------


dfsp-spirit (approved):
Thanks, and good idea to add an extra parameter for the FT behavior. I guess not everybody will want this.

PR #464, opened 2023-04-03
FIX: fix TestMetadataUsingFooof.test_channel_par

dfsp-spirit:
Fix: Determine `TestMetadataUsingFooof.test_channel_par()` unit test expectations based on whether parallel processing was *actually used* (as opposed to *requested*).

This fixes things on the MacOS runners at ESI. On GitHub CI, this is not triggered, as the tests actually run in parallel there.

Explanation:
It turned out that the test fails if parallel computation is not actually possible on the system, i.e., you set `cfg.parallel=True`, but actually sequential computation is being used (with a warning, but the test does not know about that). I solved this by adding a `log_dict` entry `used_parallel`, which should also come in handy in other situations.

Reviewer Checklist
------------------

### DIFF
--- a/syncopy/shared/computational_routine.py
+++ b/syncopy/shared/computational_routine.py
@@ -607,6 +607,9 @@ class ComputationalRoutine(ABC):
         # Create HDF5 dataset of appropriate dimension
         self.preallocate_output(out, parallel_store=parallel_store)
 
+        log_dict = {} if log_dict is None else log_dict
+        log_dict["used_parallel"] = str(parallel)
+
         if parallel:
 
             # Construct list of dicts that will be passed on to workers: in the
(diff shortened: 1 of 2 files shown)


syncopy/shared/computational_routine.py
```diff
@@ -607,6 +607,9 @@ class method (starting with the word `"compute_"`).
         # Create HDF5 dataset of appropriate dimension
         self.preallocate_output(out, parallel_store=parallel_store)
 
+        log_dict = {} if log_dict is None else log_dict
+        log_dict["used_parallel"] = str(parallel)
```
tensionhead, line 611: this is good!

syncopy/tests/test_metadata.py
```diff
@@ -395,7 +395,11 @@ def test_channel_par(self):
         spec_dt = freqanalysis(cfg, multi_chan_data, fooof_opt=fooof_opt, chan_per_worker=chan_per_worker)
 
         # How many more calls we expect due to channel parallelization.
-        calls_per_trial = int(math.ceil(num_channels / chan_per_worker))
+        used_parallel = 'used_parallel = True' in spec_dt._log
```
tensionhead, line 398: aah.. yes, makes all sense now :+1:

tensionhead (approved):

PR #481, opened 2023-04-14
CHG: Handle NaNs in preprocessing

### LINKED ISSUE #480: Handling of NaNs during FIR filtering
From the [FT documentation](https://github.com/fieldtrip/fieldtrip/blob/master/preproc/ft_preproc_bandstopfilter.m):
```
% If the data contains NaNs, these will affect the output. With an IIR
% filter, and/or with FFT-filtering, local NaNs will spread to the whole
% time series. With a FIR filter, local NaNs will spread locally, depending
% on the filter order.
```
Syncopy always uses SciPy's `fftconvolve` for FIR filters, hence even a single NaN will also spread to the whole timeseries. We should add a check in the respective FIR cF for NaNs, and if one is found switch to the slower time domain ("direct") convolution (via `np.convolve`).

Additionally, the cF should additionally return the result of the NaN check to emit a warning to the users.

tensionhead:
Changes Summary
----------------
- FIR filtering switches to slower time-domain convolutions if NaNs are present, still giving results if at least `order` (filter length) away from any NaN samples
- for IIR filtering, we can't do anything but the user gets a warning and the offending (NaN containing) trials can be accessed via `.info['nan_trials']`
- we probably could roll out this NaN reporting also elsewhere (specest, connectivity, ...)
- closes #480

Reviewer Checklist
------------------

### DIFF
--- a/syncopy/preproc/preprocessing.py
+++ b/syncopy/preproc/preprocessing.py
@@ -10,7 +10,8 @@ import numpy as np
 from syncopy import AnalogData
 from syncopy.shared.parsers import data_parser, scalar_parser, array_parser
 from syncopy.shared.tools import get_defaults, get_frontend_cfg
-from syncopy.shared.errors import SPYValueError, SPYInfo
+from syncopy.shared.errors import SPYValueError, SPYInfo, SPYWarning
+from syncopy.shared.metadata import metadata_from_hdf5_file
 from syncopy.shared.kwarg_decorators import (
     unwrap_cfg,
     unwrap_select,
@@ -348,6 +349,7 @@ def preprocessing(
     # only zscoring
     else:
         filterMethod = None
+
     # -------------------------------------------
     # Call the chosen filter ComputationalRoutine
     # -------------------------------------------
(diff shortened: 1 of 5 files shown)


syncopy/preproc/preprocessing.py
```diff
@@ -368,6 +370,21 @@ def preprocessing(
...
+                # try to also record the trial numbers
+                trl_num = key.split('__')[-1].split('_')[0]
+                nan_trials.append(int(trl_num))
+
+        if len(nan_trials) != 0:
+            msg = "Data contains NaNs! See `.info['nan_trials']` for the offending trials"
+            if filter_class == 'but':
+                msg += "\n\t\t try using a 'onepass' FIR filter of low order.."
+            SPYWarning(msg)
+        filtered.info['nan_trials'] = nan_trials
```
dfsp-spirit, line 386: Good idea to pass this on to the user.

dfsp-spirit (approved):
Thanks, that's a good solution for the problem imo. Lets see in the future whether ppl will complain about similar things in other parts of syncopy.

PR #490, opened 2023-04-26
489 redesign time property

tensionhead:
Changes Summary
----------------
- new `TimeIndexer`, works very similiar to the `TrialIndexer`
- no more costly creating the full list of all trial-time arrays for a simple `.time[tk]`
- tests required NO changes, meaning that `.time` returned a list before this change was not even required and/or tested
- minor additional performance improvements for the `__str__` method

Reviewer Checklist
------------------


dfsp-spirit (approved):
Thanks! 

At first glance it seemed a bit weird that no tests complained about this, but on a second thought I'm actually fine with it. There are tests that use the time property, but they test the functionality, not the exact datatype, which is fine with me.

PR #512, opened 2023-05-22
508 pynwb

### LINKED ISSUE #511: String representation of EventData with trialdefinition raises Exception
**Describe the bug**
String representation of EventData with trialdefinition raises Exception

**To Reproduce**
Add the following unit test in tests/test_discretedata/TestEventData:
```
 def test_str_rep_with_trials(self):
        """Test string representation of EventData with trialdefinition. Ensure that the bug with the string representation is fixed."""
        dummy = EventData(self.data, trialdefinition=self.trl)
        assert "samplerate" in str(dummy)  # The real test is that 'str(dummy)' does not raise an error.
```
Run it:
```
FAILED syncopy/tests/test_discretedata.py::TestEventData::test_str_rep_with_trials - AttributeError: 'EventData' object has no attribute '_trial_ids'. Did you mean: 'trial_ids'?
```

**Expected behavior**
Test passes.

Version: current dev at [e5107bd](https://github.com/esi-neuroscience/syncopy/commit/e5107bd3303e6786df11f1e501f2e40a7b0555eb)

dfsp-spirit:
Changes Summary
----------------
- Add support for exporting `spy.TimelockData` and `SpikeData` to NWB format
- NWB files exported from Syncopy can be read with our NWB reader (`spy.io.load_nwb`). An NWB file can contain several acquisitions (equivalent to our `AnalogData/TimeLockData`) and additionally Unit-based events (equivalent to our `SpikeData`). Therefore, a dictionary of Syncopy data objects is returned by the reader.
- Before this PR, the reader could only read acquisitions, and trials were expected to be saved in the epochs field (as opposed to the trials field). 
   * Now the reader will also read trials.
   * Support was added to also read unit-based events (SpikeData) and turn them into SpikeData objects
- For validation purposes, one can open files produced with our exporter at [nwbexplorer.opensourcebrain.org](https://nwbexplorer.opensourcebrain.org). One needs to put the files on a webserver first though. I have put demo files here:
    * NWB produced by exporting SpikeData: https://rcmd.org/tmp/spikes.nwb
    * NWB produced by exporting AnalogData: https://rcmd.org/tmp/adata3.nwb
- There is also a unit test that opens an exported SpikeData file with [pynapple](https://github.com/pynapple-org/pynapple) in `syncopy/tests/test_spyio.py`, make sure to have pynapple installed in the test environment to run it (`pip install pynapple`).

Reviewer Checklist
------------------


tensionhead (commented):
Great work sifting through the extensive NWB format specifications and building an interface with Syncopy @dfsp-spirit! 

I will push some minor code/doc-string format changes.

What was not tested is `save_nwb` with `is_raw=False`, which also did not work. I will push a minor change which makes this work and allows saving LFPs. However, I could not get saving LFPs on top of raw data to work. So basically if the `nwbfile` parameter points to an existing nwbfile (read from disc via `pynwb.NWBHDF5IO`) saving additional AnalogData into that nwb dataset did not work for me. Not sure if we want that to work at this stage though? But I guess at least testing the `save_nwb(..., nwbfile=some_nwb_file`) option OR disabling it alltogether might be a good idea?

syncopy/datatype/continuous_data.py
```diff
@@ -420,6 +422,43 @@ def multipanelplot(self, **show_kwargs):
         figax = mp_plotting.plot_AnalogData(self, **show_kwargs)
         return figax
 
+    def save_nwb(self, outpath, nwbfile=None, with_trialdefinition=True, is_raw=True):
```
tensionhead, line 425: if `nwbfile` is not `None` but points to an existing nwb file, saving did not work. Probably what you mean is starting with an empty `nwbfile` container? However this is not tested..

dfsp-spirit:
> Great work sifting through the extensive NWB format specifications and building an interface with Syncopy @dfsp-spirit!
> 
> I will push some minor code/doc-string format changes.
> 
> What was not tested is `save_nwb` with `is_raw=False`, which also did not work. I will push a minor change which makes this work and allows saving LFPs. However, I could not get saving LFPs on top of raw data to work. So basically if the `nwbfile` parameter points to an existing nwbfile (read from disc via `pynwb.NWBHDF5IO`) saving additional AnalogData into that nwb dataset did not work for me. Not sure if we want support this at this stage though? But I guess at least testing the `save_nwb(..., nwbfile=some_nwb_file`) option OR disabling it alltogether might be a good idea?

Saving several of our data objects to NWB is now possible, but it is tricky. E.g., while one can store several trial definitions with different names in the NWB trials table, there is no technical way for a reading software to know which trials belong to which dataset. I think the NWB format assumes that all data shares the same trial/epoch definition, and thus only one dataset should save its trial definition in the NWB file. I have added a unit test, `test_save_nwb_analog_2`, that illustrates how to save 2 AnalogData into one NWBfile: the first one as the raw data, the second one as an LFP. This is a corner case though imo.

syncopy/datatype/continuous_data.py
dfsp-spirit: This is not fixed and there is a unit test for it, but see my detailed comment on this topic below.

syncopy/datatype/continuous_data.py
```diff
@@ -420,6 +422,48 @@ def multipanelplot(self, **show_kwargs):
...
+        """Save AnalogData in Neurodata Without Borders (NWB) file format.
+        An NWBFile represents a single session of an experiment.
+
+        Parameters
+        ----------
+        outpath : str, path-like. Where to save the NWB file, including file name and `.nwb` extension.
```
tensionhead, line 431: The `nwbfile` parameter is missing from the doc string!

tensionhead (approved):
Great work @dfsp-spirit! This is very foundational work for Syncopy, as interoperability is key in the maze of Neuroscientific dataformats. Fingers crossed that NWB establishes itself more and more..

PR #515, opened 2023-05-24
Fix crash on interactive cleanup twice

dfsp-spirit:
Changes Summary
----------------
- insta-fix for a bug discovered by DJ: on the latest dev, running `spy.cleanup(interactive=False)` twice in a row crashes Syncopy., #516 
- cannot easily add tests for the bug to verify that it is fixed: running spy.cleanup while other tests are run in parallel may remove their temp files, and thus make other tests fail at random.

Reviewer Checklist
------------------


tensionhead (approved):
Quick and easy solution :+1:

PR #517, opened 2023-05-24
434 add plotting for spikedata

tensionhead:
Changes Summary
----------------
- `singlepanelplot` and `multipanelplot` for `SpikeData`
- can plot either indexed by unit or channel
- hotfix `_trial_ids` für DiscreteData
- form is a so called "spike raster plot":
Screenshot from 2023-05-30 17:41:25

Reviewer Checklist
------------------


dfsp-spirit (approved):
very nice plots, and good to have them plot unit instead of channel. I guess people like Katharine are more qualified than me to comment on them, but from what I see online, this is suitable. some people seem to plot a single neuron, with the trials on the y axis, but if really needed, we can add that later.

tensionhead:
Thx for the review! I'll ping @kshapcott and @Cumol directly, adding the trial axis is not a biggy I think.

dfsp-spirit:
Since there have been no replies yet, I would now merge this so it can be used, e.g., for the docs. If we need some minor changes (e.g., add labels to multi-channel plots) or additions (like trial axis), we can add them later.

Cumol:
sorry for the lack of response. looks good! will give it a try in the next week or two when I am touching spike data again.

Another typical plot is to plot one unit over multiple trials. So the x-axis would be time, y axis would be trials, and the subplots would be for each unit.

What do yu say @KatharineShapcott ?

KatharineShapcott:
Yes I agree, x-axis time, y axis trials.

Issue #491, opened 2023-05-02
Better MNE-Integration: live conversion between our and their data classes


dfsp-spirit:
We encourage use of both Syncopy and MNE, and do not want to reinvent the wheel by building stuff that is already available in MNE. To make working with both software systems easier, it would be nice to make it easier to convert data between MNE and Syncopy. 

**Do you have a solution in mind? Please elaborate**
Would be good to add some wrapper methods for easy conversion between our dataclasses (like AnalogData) and the MNE data classes (Raw, Epcohed, ...).

**Additional context**
Some examples where Syncopy users may want to switch to MNE is for stuff like ICA and source localization.

PR #527, opened 2023-06-07
491 Better MNE integration

dfsp-spirit:
Changes Summary
----------------
- improve MNE integration, see #491 
- add some helper functions for converting between:
   * Syncopy and MNE raw data (raw data is AnalogData without trial definition): `raw_mne_to_adata`, `raw_adata_to_mne`
   * Syncopy and MNE epoched data (epoched data is timelocked, with trial definition)
- add member function `ìs_time_locked()` to continuousdata
- These functions use the MNE data classes (e.g., `mne.io.RawArray`), and only work if MNE Python (package `mne`) is installed. MNE is huge and has many dependencies, so we do NOT add it as a syncopy dependency. If users want to use these functions, they will have to manually install MNE. (If they do not have MNE installed, the functions will be useless anyways).

Reviewer Checklist
------------------


syncopy/io/mne_conv.py
```diff
@@ -0,0 +1,134 @@
...
+    n_times = ea.get_data().shape[2]
+    spy_data = np.zeros((n_times * n_epochs, n_channels), dtype=np.float32)
+
+    for chan_idx in range(n_channels):
+        spy_data[:, chan_idx] = ea.get_data()[:,chan_idx,:].flatten()
+
+    tldata = spy.AnalogData(data=spy_data, samplerate=ea.info['sfreq'], channel=ea.ch_names)
+
+    nSamples = n_times
+    trldef = np.vstack([np.arange(0, nSamples * n_epochs, nSamples),
```
tensionhead, line 130: Yep, very concise 👍

syncopy/io/mne_conv.py
```diff
@@ -0,0 +1,134 @@
...
+                        np.arange(0, nSamples * n_epochs, nSamples) + nSamples,
+                        np.ones(n_epochs) * -1]).T
```
tensionhead, line 132: so here all offsets are set to `-1`? This determines the `0` of the time-axis, not sure how this is encoded in MNE actually..

syncopy/tests/test_mne_conv.py
```diff
@@ -0,0 +1,158 @@
...
+
+        # Check dimensions
+        n_times = epoched.get_data().shape[2]
+        assert n_times == self.numSamples
+        n_epochs = epoched.get_data().shape[0]
+        assert n_epochs == self.numTrials
+        n_channels = epoched.get_data().shape[1]
+
+        assert n_times == tldata.trials[0].shape[0]
+        assert n_epochs == len(tldata.trials)
```
tensionhead, line 85: `epochs = trials`.. jeez all this nomenclature still is quite confusing

syncopy/tests/test_mne_conv.py
```diff
@@ -0,0 +1,158 @@
...
+        n_channels = epoched.get_data().shape[1]
+
+        assert n_times == adata.trials[0].shape[0]
+        assert n_epochs == len(adata.trials)
+        assert n_channels == len(adata.channel)
+
+    @skip_no_mne
+    def test_mne_epoched_to_AnalogData(self):
+        """
+        Test conversion of spy.AnalogData that is time locked to mne.EpochsArray.
```
tensionhead, line 119: Here you probably copy-pasted from the doc string above? It should say `..conversion of mne.EpochsArray to spy.AnalogData` ?!

tensionhead (commented):
Pretty cool, looks fine to me. I just had some minor comments, please see below. 
It is naturally a bit confusing with the *timelocked spy.AnalogData* and then our *spy.TimelockData*. The latter I would consider a special case of limited use, as we discussed (can't serve as input for `freqanalysis` or any other frontend as of now). The former now has a formal standing thanks to the new `is_time_locked` property 👍

syncopy/datatype/continuous_data.py
```diff
@@ -67,6 +67,17 @@ def data(self, inData):
 
         if inData is None:
             return
+        
+    @property
+    def is_time_locked(self):
```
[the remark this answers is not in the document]
tensionhead: Fixed with bf090d8799a5307adad103fc2e8886083cbf149f, hopefully this now catches all cases..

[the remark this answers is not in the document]
dfsp-spirit: Thanks!

syncopy/io/mne_conv.py
```diff
@@ -0,0 +1,134 @@
...
+    ar : `mne.io.RawArray` instance
+    """
+
+    try:
+        import mne
+    except ImportError:
+        raise ImportError("MNE Python not installed, but package 'mne' is required for this function.")
+    data_parser(adata, varname="adata", dataclass="AnalogData")
+    info = mne.io.meas_info.create_info(list(adata.channel), adata.samplerate, ch_types='misc')
+    ar = mne.io.RawArray((adata.data[()]).T, info)
```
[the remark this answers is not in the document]
dfsp-spirit: Yes, for the raw data, it's always easy.

syncopy/datatype/continuous_data.py
```diff
@@ -790,12 +801,11 @@ def trialdefinition(self, trldef):
...
         # now check for additional conditions
 
-        # FIXME: not clear, is timelocked data to be expected
-        # to have same offsets?!
-        # if not np.unique(trldef[:, 2]).size == 1:
-        #     lgl = "equal offsets for timelocked data"
-        #     act = "different offsets"
-        #     raise SPYValueError(lgl, varname="trialdefinition", actual=act)
+        # Timelocked data should have same offset for all trials
+        if not np.unique(trldef[:, 2]).size == 1:
```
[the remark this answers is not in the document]
dfsp-spirit: thanks for changing

syncopy/io/mne_conv.py
```diff
@@ -0,0 +1,134 @@
...
+
+    Returns
+    -------
+    epochs : `mne.EpochsArray` instance
+    """
+    try:
+        import mne
+    except ImportError:
+        raise ImportError("MNE Python not installed, but package 'mne' is required for this function.")
+    assert type(tldata) == spy.AnalogData or type(tldata) == spy.TimeLockData, "Invalid input: tldata must be of type AnalogData or TimeLockData."
```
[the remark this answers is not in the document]
dfsp-spirit: fixed in 83804d9b

syncopy/tests/test_mne_conv.py
dfsp-spirit: myunderstanding: for MNE, there are trials (the short recording snippets, one for each repetition of a task) and epochs (subsets of recordings, which  are timelocked with respect to some event during the trial, i.e., the event happens at the same time relative to the start of the epoch, for all epochs).

In MNE, people seem to use only epochs during the analysis in many cases, as trials are something from the raw data domain that is typically not of relevance for the analysis. syncopy does not have epochs, so we have to store epochs as trials.

syncopy/tests/test_mne_conv.py
dfsp-spirit: fixed

syncopy/tests/test_mne_conv.py
tensionhead: Uh, ok that's even more confusing. A proper definition of what FT/Syncopy considers a trial would help for sure. Just not overly confident that we are the perfect ppl to write it 😅

dfsp-spirit: MNE has a definition of what an Epoch is (or, what their EpochsArray datastructure models) at https://mne.tools/stable/auto_tutorials/epochs/10_epochs_overview.html

Quoting the relevant part:

[Epochs](https://mne.tools/stable/generated/mne.Epochs.html#mne.Epochs) objects are a data structure for representing and analyzing equal-duration chunks of the EEG/MEG signal. [Epochs](https://mne.tools/stable/generated/mne.Epochs.html#mne.Epochs) are most often used to represent data that is time-locked to repeated experimental events (such as stimulus onsets or subject button presses), but can also be used for storing sequential or overlapping frames of a continuous signal (e.g., for analysis of resting-state activity; see [Making equally-spaced Events arrays](https://mne.tools/stable/auto_tutorials/raw/20_event_arrays.html#fixed-length-events)). Inside an [Epochs](https://mne.tools/stable/generated/mne.Epochs.html#mne.Epochs) object, the data are stored in an [array](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.html#numpy.ndarray) of shape (n_epochs, n_channels, n_times).

dfsp-spirit: I will try to write something, and let Diljit and Mohsen double check it.

syncopy/io/mne_conv.py
dfsp-spirit: I think because the imported data is epoched already, we want our trials to span the entire input data. Should this be `np.zeros(n_epochs)` then instad of `np.ones(n_epochs) * -1])`?

dfsp-spirit:
I have also added links to the new functions in the HTML docs, so people will actually find them.

syncopy/io/mne_conv.py
```diff
@@ -0,0 +1,139 @@
...
+
+    This function requires MNE Python (package 'mne') and will raise an `ImportError` if it is not installed.
+
+    Parameters
+    ----------
+    ea : `mne.EpochsArray` instance
+
+    Returns
+    -------
+    tldata : `syncopy.AnalogData` instance. The trial definition will be set to the MNW epochs, and it is guranteed that the data is time-locked.
```
tensionhead, line 114: MNE instead of MNW?!

syncopy/io/mne_conv.py
tensionhead: I think `offset = -ea.tmin * samplerate` would do the trick, then just  `np.ones(n_epochs) * offset` as last column in the trialdef.

syncopy/io/mne_conv.py
```diff
@@ -0,0 +1,139 @@
...
+    Returns
+    -------
+    adata : `syncopy.AnalogData` instance, with no trial definition (singl-trial data).
+    """
+    try:
+        import mne
+    except ImportError:
+        raise ImportError("MNE Python not installed, but package 'mne' is required for this function.")
+
+    assert type(ar) == mne.io.RawArray, "Invalid input: ar must be of type mne.io.RawArray."
```
tensionhead, line 61: Also here better to use `SPYTypeError`

dfsp-spirit: fixed in b6352542

syncopy/io/mne_conv.py
dfsp-spirit: fixed in e5569446

tensionhead (approved):
Looks all good, thx @dfsp-spirit !

PR #528, opened 2023-06-11
Documentation sprint23

tensionhead:
Changes Summary
----------------
- Shortened and partial rewriting of the Quickstart
- new Tutorials: Connectivity, Resampling and Preprocessing
- adaptations to the new `spy.synthdata` module
- new global function: `spy.set_loglevel`


dfsp-spirit (approved):
This is an impressive improvement to our documentation, thanks heaps. It's also way more than just Syncopy docs, one can even see it as a combination of syncopy docs and an introduction to signal analyis, thanks to all the background information contained in it. I think this is the best form of documentation: it now only shows you how to perform an analysis, but also explains the results you see and why you are sometimes seeing things one may not expect. 

I directly changed some minor typos, other than that I have no complaints for now.

PR #565, opened 2023-08-24
563 add channelcmb parameter to connectivityanalysis frontend

tensionhead:
Changes Summary
----------------
- new parameter `channelcmb=[senders, receivers]` for `spy.connectivityanalysis`
- `selectdata` now allow non-contiguous and unorderd `channel_i` and `channel_j` selections, there was no reason apparent why this was not allowed before (I did not fiddle with any selection mechanics inside the `Selector`)

Now users can fine tune the channel-pairs for which to compute the connectivity measure as in FieldTrip:
 ```python
senders = [0, 3] 
receivers = [1, 5, 8]
spy.connectivityanalysis(..., channelcmb=[senders, receivers])
```
will compute only 6 pairs: (01, 05, 08, 31, 35, 38) and the resulting `CrossSpectralData` **is rectangular in the channel axes** with ` channel_i=senders` and `channel_j=receivers`. 

Note: the old way of doing a channel selection on the input:
```python
spy.connectivityanalysis(..., select={'channel': [0, 1, 3, 5, 8]})
```
would instead compute 20 pairs (+diagonal): the full dyadic (or tensor) product of those 5 channels.

#### Granger causality
- computes **all pairs individually**, so in the example above this would be 6 spectral factorizations of 2 x 2 matrices. The result gets collected into a single 1 x nFreq x 2 x 3 `CrossSpectralData` object. The factorization algorithm is most stable when only computing pairs, however for many trials/few channels it would be much more effective to compute more pairs at once. 

#### CSD/PPC
- the sender/receiver pairs get directly indexed within the cF (so the numpy arrays), and the single trial cross-spectra have rectangular shape. It was straightforward inputting this into the PPC algorithm, to compute the PPC only for the desired pairs.

#### Coherence
- rectangular single trial cross-spectra are not an easy option here, as the individual auto-spectra (diagonal in the full dyadic case) are needed for normalization. However, since computation is extremely cheap, all what is done is a simple post-selection of the `channel_i/senders` `channel_j/receivers` combinations which now is possible.

### Limitations
- only for spectral methods `('coh', 'csd', 'ppc', 'granger')`
- works only for spectral input (`SpectralData`) as re-writing the whole implicit `freqanalysis` part when inputting `AnalogData` into `connectivityanalysis` would require some bigger refactoring (design is actually bad for historic reasons)
- general inplace selections on the input are still fine of course, but channel selections don't work together with `channelcmb`

Reviewer Checklist
------------------

### DIFF
--- a/syncopy/connectivity/connectivity_analysis.py
+++ b/syncopy/connectivity/connectivity_analysis.py
@@ -385,7 +385,7 @@ def connectivityanalysis(
         st_dimord = CrossCovariance.dimord
 
     # all these methods need the single trial cross spectra
-    # we just have to sort out if we need an mtmfft first
+    # we just have to sort out if we need an mtmfft first (for AnalogData input)
     elif method in ["csd", "coh", "ppc", "granger"]:
         nTrials = len(data.trials)
         if nTrials == 1:
@@ -405,7 +405,7 @@ def connectivityanalysis(
             # the actual number of samples in case of later padding
             nSamples = process_padding(pad, lenTrials, data.samplerate)
 
-            check_effective_parameters(CrossSpectra, defaults, lcls, besides=("jackknife",))
+            check_effective_parameters(CrossSpectra, defaults, lcls, besides=("jackknife", "channelcmb"))
 
             st_compRoutine, st_dimord = cross_spectra(
                 data,
@@ -445,11 +445,12 @@ def connectivityanalysis(
                 SpectralDyadicProduct,
                 defaults,
                 lcls,
-                besides=("jackknife"),
+                besides=("jackknife", "channelcmb"),
             )
 
-            # sanity checks done above
-            if channelcmb is not None:
+            # sanity checks for channelcmb done above, truly rectangular matrix
+            # operations only meaningful for PPC or single trial cross-spectra
+            if channelcmb is not None and method in ['ppc', 'csd']:
                 sender, receiver = channelcmb
 
                 # save current selection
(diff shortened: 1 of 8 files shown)


dfsp-spirit:
Wow, great to hear that you did not have to mess with the Selector logic at all to allow non-contiguous and unorderd channel_i and channel_j selections.

syncopy/connectivity/connectivity_analysis.py
```diff
@@ -543,16 +666,71 @@ def connectivityanalysis(
...
+                    av_compRoutine.pre_check()  # make sure we got a trial_average
+                    av_compRoutine.compute(st_out, pair_out, parallel=kwargs.get("parallel"), log_dict=log_dict)
+
+                    # write result, idx1/idx2 directly encode positions in result matrix
+                    with h5py.File(fname, "r+") as h5file:
+                        dset = h5file['data']
+
+                        # only direction sender(ch1) -> receiver(ch2)
+                        dset[0, :, idx1, idx2] = pair_out.data[0, :, 0, 1]
+
```
dfsp-spirit, line 713: nice

syncopy/datatype/selector.py
```diff
@@ -735,16 +735,14 @@ def _selection_setter(self, data, select, selectkey):
...
-                        "channel_j",
-                    ]:
-                        if len(idxList) > 1:
-                            err = "Unordered (low to high) or non-contiguous multi-channel-pair selections not supported"
-                            raise NotImplementedError(err)
-                        idxList = idxList[0]
+                            "channel_i",
+                            "channel_j"]:
+
+                        # why only for CrossSpectralData a 1-element
```
dfsp-spirit, line 742: beats me

dfsp-spirit (approved):
Thanks a lot, this will make the code for the paper look a lot better.

